PR #115, opened 2026-03-10
dockerfile for sable call, org repo cleanup

melogale:
I bring in the new sable call as a submodule!
And adjust the dockerfile and builds accordingly.
(Github workflows almost done.)


hazre:
No submodule please! Use git urls to reference it in package.json instead

melogale:
Can do.

PR #129, opened 2026-03-10
feat: Implement an UI for commands

Rosy-iso:
### Description

Adds an interface that replicates the behavior of the following commands:

- [x] /myroomavatar
- [x] /myroomnick
- [x] /color
- [x] /scolor
- [x] /pronoun
- [x] /spronoun
- [x] /font
- [x] /sfont

The original commands will still remain and be usable however.
(In fact the interface calls the commands in the background, so the commands should not be changed.)

#### Type of change

### Checklist:


7w1:
Only issue I see is that there's no feedback for if the user has permission to change their pronouns/color/font in a space. Could disable those fields similar based on whether or not the user has permission (look at the code for the toggles and disable if the toggles are disabled, or similar).

Rosy-iso:
oh right, i forgot those could be disabled
I'll make them disable when the user doesn't have permission

Rosy-iso:
should be all good now

7w1 (approved):

PR #232, opened 2026-03-12
Optimize sliding sync with progressive loading

Just-Insane:
- Implement 100-room chunk loading for sliding sync
- Add 100ms delay before active room subscription  
- Fix timeline range management for better scroll behavior
- Optimize timeline limits based on device capabilities (50-150 events)


Just-Insane:
Still some slight jumpiness with when loading but this cuts down a majority of it now

Issue #216, opened 2026-03-12
Reaction events increase unread badge count


Ethanol-0:
### Describe the bug

If someone reacts to a post, the channel gets marked as unread. See for example the Sable Announcement channel which keeps showing as unread because people react to the posts.

### Reproduction

1. mark channel as read
2. have someone else react to a message
3. watch channel now marked as unread with no new text message

### Expected behavior

reactions should probably not count as unread messages

### Platform and versions

```shell
- OS: Linux x86_64
- Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Sable/0.0.1 Chrome/144.0.7559.177 Electron/40.6.0 Safari/537.36
- Sable: v1.7.0 (f50de89)
```

### Additional context

using Sable Desktop AppImage

DidiDidi129:
Experiencing as well

leamikmik:
I think this happens because of [this line](https://github.com/SableClient/Sable/blob/dev/src/app/utils/room.ts#L215) (introduced in 976d2e8) since it makes all reactions a notification event. I think it would be best for only reactions to your messages be displayed, but I would prefer the Discord style (you get a push notification, not a unread)

Just-Insane:
Should be fixed by https://github.com/SableClient/Sable/pull/233/changes#diff-fc3ae8de2ac33e76cab370fad07a25997dc93437047285745b76def8b15c3696R231

7w1:
Fixed in #233

Issue #167, opened 2026-03-10
Valid message is rendered as broken


dozro:
### Describe the bug

The [following message](https://matrix.to/#/!a6sXbRuOyyc7MKutmy:sable.moe/$QUbSte-uYnuoBtDFcW7Ye4Z1o31k9sz7JvR7hA0O1ac?via=matrix.org&via=unredacted.org&via=blossomdevs.com&via=fairchat.eu) is rendered as broken message

```json
{
  "content": {
    "body": "",
    "msgtype": "m.text"
  },
  "origin_server_ts": 1773182126392,
  "sender": "@rye:fairchat.eu",
  "type": "m.room.message",
  "unsigned": {
    "membership": "join",
    "age": 337,
    "transaction_id": "m1773182126184.46"
  },
  "event_id": "$QUbSte-uYnuoBtDFcW7Ye4Z1o31k9sz7JvR7hA0O1ac",
  "room_id": "!a6sXbRuOyyc7MKutmy:sable.moe"
}
```

<img width="230" height="44" alt="Image" src="https://github.com/user-attachments/assets/1d7458af-361f-4159-9dbf-0a7fca0b292c" />

it should be a valid message, maybe display it as "empty message" or something, but not as broken

### Platform and versions

```shell
Sable: v1.6.0 (d9a8dae)
```

Rosy-iso:
what does the spec say

dozro:
good question, i will need to look that up
@nexy7574 meant in the discussion on matrix yesterday that an empty body should be allowed, while an absent body shouldn’t 

Edit: As far as I read the spec an empty body is allowed. It's not really ideal, but it is allowed.

leamikmik:
Also getting this error on a message with an empty body but a non-empty formatted_body

timedoutuk:
The spec mandates that `body` is present, not that it is populated. For example, an image without a caption will have `"body": ""` to denote that there is no caption. Regular text messages can also be empty if the user chooses not to send any text.

dozro:
closed with https://github.com/SableClient/Sable/pull/234

PR #234, opened 2026-03-12
Fix reaction clicks, zoom persistence, and empty message rendering

### LINKED ISSUE #215: Page zoom doesn't affect message text size
### Describe the bug

Increasing the "page zoom" percentage in appearance settings will increase many elements of the UI, but curiously does not affect the text size in messages. This works how I expect it to in Cinny. I can't find any other way to increase the message text size to work around this.

### Reproduction

1. Open settings, go to Appearance, set page zoom to something higher like 150% or 200%.
2. Go into a text room.
3. Text hasn't changed size.

### Expected behavior

Text should also scale up by the same fraction.

### Platform and versions

```shell
- Firefox 140.8.0esr (64-bit)
- Connecting to app.stable.moe
```

### Additional context

_No response_

Just-Insane:
Fixes #215, #167

- Fix reaction click events not firing correctly
- Persist page zoom level across sessions  
- Handle empty message rendering edge cases

### DIFF
--- a/.changeset/fix-ui-bugs.md
+++ b/.changeset/fix-ui-bugs.md
@@ -1,0 +1,5 @@
+---
+'default': patch
+---
+
+Fix reaction clicks, zoom persistence, and empty message rendering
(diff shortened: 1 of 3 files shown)


dozro:
Does this actually fix https://github.com/SableClient/Sable/issues/216?

Just-Insane:
Ah, looks like that's actually here: https://github.com/SableClient/Sable/pull/233/changes#diff-fc3ae8de2ac33e76cab370fad07a25997dc93437047285745b76def8b15c3696R231

PR #282, opened 2026-03-15
feat: edit file description

nushea:
### Description

Added the ability to edit one's file descriptions

Additionally changed the UX for images and video rendering. 
Now Idle or blurred images/videos can be loaded by clicking anywhere on the image/video, not just in the small buttons.
Changed the behavior of videos to not autoplay on load as previously when you had a Watch option it was usually blurred due to it not being fully loaded, not offering you the chance to see the thumbnail before starting the video.

(view of the editing of a file description with the inline setting)
image

(hovering an image/video with your mouse while it is unblurred now has a button to blur it)
image

(hovering over a blurred image, with the text changed to Spoilered as now it will display the spoiler reason within that false button, presumably, i have never seen an image with a spoiler reason but the logic is just how it worked previously with tooltips but more clearly labeled)
image

#### Type of change

### Checklist:

### DIFF
--- a/src/app/components/message/content/ImageContent.tsx
+++ b/src/app/components/message/content/ImageContent.tsx
@@ -209,37 +209,24 @@ export const ImageContent = as<'div', ImageContentProps>(
               }
             }}
           >
-            {typeof spoilerReason === 'string' && spoilerReason.length > 0 ? (
-              <Chip
-                variant="Secondary"
-                radii="Pill"
-                size="500"
-                outlined
-                onClick={() => {
-                  setBlurred(false);
-                  if (srcState.status === AsyncStatus.Idle) {
-                    loadSrc();
-                  }
-                }}
-              >
-                <Text size="B300">Spoiler reason: {spoilerReason}</Text>
-              </Chip>
-            ) : (
-              <Chip
-                variant="Secondary"
-                radii="Pill"
-                size="500"
-                outlined
-                onClick={() => {
-                  setBlurred(false);
-                  if (srcState.status === AsyncStatus.Idle) {
-                    loadSrc();
-                  }
-                }}
-              >
-                <Text size="B300">Spoilered</Text>
-              </Chip>
-            )}
+            <Chip
+              variant="Secondary"
+              radii="Pill"
+              size="500"
+              outlined
+              onClick={() => {
+                setBlurred(false);
+                if (srcState.status === AsyncStatus.Idle) {
+                  loadSrc();
+                }
+              }}
+            >
+              <Text size="B300">
+                {typeof spoilerReason === 'string' && spoilerReason.length > 0
+                  ? `Spoiler reason: ${spoilerReason}`
+                  : `Spoilered`}
+              </Text>
+            </Chip>
           </Box>
         )}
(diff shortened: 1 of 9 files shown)


src/app/components/message/content/ImageContent.tsx
```diff
@@ -250,6 +282,26 @@ export const ImageContent = as<'div', ImageContentProps>(
...
         )}
+        {isHovered && (
+          <Box style={{ padding: config.space.S200, right: 0, position: 'absolute' }}>
+            <Menu style={{ padding: config.space.S0 }}>
+              <MenuItem
+                size="300"
+                after={<Icon size="200" src={blurred ? Icons.Eye : Icons.EyeBlind} />}
+                radii="300"
+                fill="Soft"
+                variant="Secondary"
```
dozro, line 280: maybe add something like this for accessibility sake, second try on this comment sorry -- didn't know i had to include the line already in the code

```suggestion
                variant="Secondary"
                title={blurred ? 'Reveal Image' : 'Hide Image'}
```

src/app/components/message/content/VideoContent.tsx
```diff
@@ -217,6 +252,26 @@ export const VideoContent = as<'div', VideoContentProps>(
...
         )}
+        {isHovered && (
+          <Box style={{ padding: config.space.S200, right: 0, position: 'absolute' }}>
+            <Menu style={{ padding: config.space.S0 }}>
+              <MenuItem
+                size="300"
+                after={<Icon size="200" src={blurred ? Icons.Eye : Icons.EyeBlind} />}
+                radii="300"
+                fill="Soft"
+                variant="Secondary"
```
dozro, line 250: similar to the image one :)
always helpful to give the use cues what they're doing

```suggestion
                variant="Secondary"
                title={blurred ? 'Reveal Video' : 'Hide Video'}
```

dozro (changes requested):

dozro:
i noticed that your pr changes how the editor is displayed

**the dev branch**:

Screenshot 2026-03-16 at 00 07 57

**your change**:

Screenshot 2026-03-16 at 00 09 45

is that intentional?

nushea:
> i noticed that your pr changes how the editor is displayed
> 
> **the dev branch**:
> Screenshot 2026-03-16 at 00 07 57
> 
> **your change**:
> Screenshot 2026-03-16 at 00 09 45
> 
> is that intentional?

yes this is caused by the one line change in src/app/styles/overrides/Composer.css.ts that was intended to fix  min-width 100% causing the inline editor to spill, when i made that fix i was using an older version of sable however and it seems to have been fixed by something else. This change was initially made on a branch i have since deleted, i am going to revert that one line change because i cannot recreate the spilling right now

(like it now works well w my tests with the line reverted per this example)
image

src/app/components/message/content/VideoContent.tsx
nushea: yup i have added the titles to the buttons this is indeed something very important for accessibility, should keep accessibility more in the front of my mind sometimes

dozro (approved):

dozro:
Should be okay now. Thanks :)

PR #303, opened 2026-03-16
fix(timeline): prevent emoji color bleed in read receipt display names

### LINKED ISSUE #267: emoji in display name causes discoloration in read receipts for following user names
### Describe the bug

so it seems that if one has an emoji in their display name, it causes the display name of the following users to change its color to the color of the emoji, however that even happens…

Image

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
- OS: iPad
- Browser: Mozilla/5.0 (iPad; CPU OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.3 Mobile/15E148 Safari/604.1
- Sable: v1.7.0 (f50de89)
```

### Additional context

_No response_

Just-Insane:
## Summary

Fixes #267 — On Safari/iOS, colorful emoji in user display names (e.g. `🔴 Alice`) bled their WebKit text fill color into adjacent text spans in the "following the conversation" indicator at the bottom of the timeline.

### Root cause

WebKit's implementation of `-webkit-text-fill-color` propagates from emoji characters in a `<b>` tag into sibling elements under certain conditions. When a display name containing emoji was rendered as `<b>🔴 Name</b>`, the emoji's color fill propagated to adjacent `<Text>` spans like " and 2 others are following".

### Fix

Added `WebkitTextFillColor: 'inherit'` to all `<b>` name tags in `RoomViewFollowing.tsx` to contain the WebKit fill color within the tag, and `WebkitTextFillColor: 'currentColor'` to the adjacent `<Text as="span">` elements to restore the expected text color explicitly.

Applied to all 4 user-count rendering branches (1, 2, 3, and >3 users).

### DIFF
--- a/src/app/features/room/RoomViewFollowing.tsx
+++ b/src/app/features/room/RoomViewFollowing.tsx
@@ -89,56 +89,56 @@ export const RoomViewFollowing = as<'div', RoomViewFollowingProps>(
               <Text size="T300" truncate>
                 {names.length === 1 && (
                   <>
-                    <b>{names[0]}</b>
-                    <Text as="span" size="Inherit" priority="300">
+                    <b style={{ WebkitTextFillColor: 'inherit' }}>{names[0]}</b>
+                    <Text as="span" size="Inherit" priority="300" style={{ WebkitTextFillColor: 'currentColor' }}>
                       {' is following the conversation.'}
                     </Text>
                   </>
                 )}
                 {names.length === 2 && (
                   <>
-                    <b>{names[0]}</b>
-                    <Text as="span" size="Inherit" priority="300">
+                    <b style={{ WebkitTextFillColor: 'inherit' }}>{names[0]}</b>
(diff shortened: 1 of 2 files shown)


src/app/features/room/RoomViewFollowing.tsx
```diff
@@ -89,56 +89,106 @@ export const RoomViewFollowing = as<'div', RoomViewFollowingProps>(
               <Text size="T300" truncate>
                 {names.length === 1 && (
                   <>
-                    <b>{names[0]}</b>
-                    <Text as="span" size="Inherit" priority="300">
+                    <b style={{ WebkitTextFillColor: 'inherit' }}>{names[0]}</b>
```
dozro, line 92: is there a reason why we mix raw html `<b>` and react components `<Text>` here, i always wondered lol

PR #304, opened 2026-03-16
feat(search): intercept Ctrl+F / Cmd+F to open Sable search

### LINKED ISSUE #179: Override ctrl + f to show the search window instead.
### Describe the problem

right now ctrl + f opens normal browser search.

It should be context aware and open the search window instead. Like ctrl + f on a room should start the search window with that room selected.

### Alternatives considered

_No response_

### Additional context

_No response_

Just-Insane:
## Summary

Fixes #179 — Pressing `Ctrl+F` / `Cmd+F` previously opened the browser's native find-in-page dialog. It now opens Sable's search modal instead.

### Change

In `SearchModalRenderer`, the keyboard shortcut handler was extended to also match `mod+f` in addition to the existing `mod+k`. Both shortcuts now:
- `preventDefault()` the event (suppressing browser find-in-page)
- Toggle the search modal open/closed
- Skip opening if another overlay (portal) is already visible

The hint text in the search modal footer was updated to show both `Ctrl+K / Ctrl+F` (or `⌘+K / ⌘+F` on macOS).

### DIFF
--- a/.changeset/feat-ctrl-f-search.md
+++ b/.changeset/feat-ctrl-f-search.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Add Ctrl+F / Cmd+F keyboard shortcut to open Sable search instead of browser find-in-page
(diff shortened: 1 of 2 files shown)


Just-Insane:
Tested in browser, works as expected.

PR #342, opened 2026-03-17
fix(autocomplete): Enter always selects first item; first item not highlighted on open

Just-Insane:
## Problem

Two related bugs in the autocomplete menu:

1. **Enter always triggered the first item**: `RoomInput` was querying `button[data-selected]` (truthy selector — matches any button with the attribute, regardless of value), so `button[data-selected="false"]` was selected just as readily as `button[data-selected="true"]`. Pressing Enter always clicked the first button no matter what item the user had navigated to.

2. **First item not highlighted on open**: `AutocompleteMenu`'s `useLayoutEffect` had `[selectedIndex]` in its dep array. When `selectedIndex` was already `0` (the initial value) and new buttons appeared (async search results loading), the effect didn't re-run because `selectedIndex` hadn't changed — so no button ever got `data-selected="true"` stamped until the user pressed an arrow key.

## Fix

**`RoomInput.tsx`**: Change the querySelector to `button[data-selected="true"]` so only the actually-selected item is targeted.

**`AutocompleteMenu.tsx`**: Remove the dep array entirely (run after every render). The `setSelectedIndex` call inside is guarded by a button-count comparison, so it only fires when the item list changes — no infinite update loop. A `prevButtonCountRef` tracks this across renders. This ensures newly-loaded buttons are stamped immediately.

## Tests

`AutocompleteMenu.test.tsx` (new file) — 7 tests covering:
- Initial render stamps first button `data-selected="true"`
- Async result arrival stamps buttons correctly (the actual bug regression test)
- Selection resets to first item on list change
- Arrow key navigation forward/backward
- Clamping at list boundaries

### DIFF
--- a/.changeset/fix-autocomplete-selection.md
+++ b/.changeset/fix-autocomplete-selection.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix autocomplete Enter key always selecting the first item and the first item not being highlighted on open.
--- a/src/app/components/editor/autocomplete/AutocompleteMenu.test.tsx
+++ b/src/app/components/editor/autocomplete/AutocompleteMenu.test.tsx
@@ -1,0 +1,193 @@
+import { type ReactNode } from 'react';
+import { describe, it, expect, vi } from 'vitest';
+import { render, act } from '@testing-library/react';
+import {
+  AutocompleteMenu,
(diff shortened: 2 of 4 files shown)


Just-Insane:
Tested as working in the PR Preview:

- [X] mentions
- [X] emoji

7w1 (approved):

PR #343, opened 2026-03-17
fix(threads): thread drawer empty with classic sync

Just-Insane:
## Problem

When using **classic sync** (non-sliding-sync), opening a thread drawer showed no messages even though replies existed in the room.

**Root cause**: Classic sync calls `room.createThread(id, root, [], false)` with an empty `initialEvents` array. The SDK creates the Thread object with only the root event in `thread.events`. The old code checked `if (fromThread.length > 0)` on the *unfiltered* array — which was truthy (length=1, the root) — and returned the filtered result (empty after removing the root), bypassing the live-timeline fallback entirely.

## Fix

**Filter first, then check**: Compute `filteredFromThread` and only skip the fallback if *that* array is non-empty. If the Thread object exists but has no replies (classic-sync case), fall through to scan `getUnfilteredTimelineSet().getLiveTimeline()`.

The logic is extracted into an exported `getThreadReplyEvents(room, threadRootId)` helper for testability.

A backfill `useEffect` is also added: when the drawer opens and the Thread object is empty, it calls `thread.addEvents()` with the live-timeline events so the authoritative Thread source is populated for subsequent renders (receipts, reactions, etc.).

## Tests

`ThreadDrawer.test.ts` (new file) — 6 tests covering:
- Normal case: Thread object has replies → returns them minus root/reactions
- Classic-sync regression: Thread has only the root → falls back to live timeline
- No Thread object at all → falls back to live timeline
- Cross-thread filtering: ignores events from other threads
- Empty result when neither source has replies

### DIFF
--- a/.changeset/fix-thread-classic-sync.md
+++ b/.changeset/fix-thread-classic-sync.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix thread drawer showing no messages when using classic sync.
--- a/src/app/features/room/ThreadDrawer.test.ts
+++ b/src/app/features/room/ThreadDrawer.test.ts
@@ -1,0 +1,155 @@
+/**
+ * Unit tests for getThreadReplyEvents — the function that decides which events
+ * to display as replies in the thread drawer.
+ *
+ * The bug these tests cover (classic-sync empty thread):
+ *   room.createThread(id, root, [], false) creates a Thread whose .events array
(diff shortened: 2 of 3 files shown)


Just-Insane:
Tested as working in the PR Preview - with classic sync enabled, threads now populate correctly.

dozro (approved):

7w1 (approved):

PR #344, opened 2026-03-17
fix(threads): reduce dead space around root message in thread drawer

Just-Insane:
## Problem

The root message section in the thread drawer had excessive padding (`S400` top / `S200` bottom), leaving a large gap above short root messages and pushing replies further down the screen.

## Fix

Reduce the padding to `S200` top / `S100` bottom so the root message sits closer to the header and the replies section. The `maxHeight: 200px` constraint and `Scroll` wrapper are unchanged — long root messages still scroll correctly.

### DIFF
--- a/.changeset/fix-thread-ui-spacing.md
+++ b/.changeset/fix-thread-ui-spacing.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Reduce dead space around the root message in the thread drawer.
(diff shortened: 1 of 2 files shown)


Just-Insane:
Dead space in thread header is slightly reduced, though could be reduced further:

image

dozro (approved):

7w1 (approved):

PR #348, opened 2026-03-17
fix: tighten sliding sync memory management

Just-Insane:
## Summary

A series of improvements to `SlidingSyncManager` memory handling, worked through in four steps.

### 1. Stop polling loop on `dispose()`

`SlidingSyncManager.dispose()` removed event listeners but never called `this.slidingSync.stop()`. The SDK's internal HTTP polling loop kept running indefinitely after the client was "disposed", leaking network connections and blocking garbage collection.

**Fix:** `dispose()` now calls `this.slidingSync.stop()` before removing listeners.

### 2. Remove adaptive timeline-limit logic

`SlidingSyncManager` previously adjusted per-room timeline limits dynamically based on connection quality (`ACTIVE_ROOM_TIMELINE_LIMIT_LOW` / `HIGH`). This complexity caused problems (spurious scroll resets on bulk-loads) and the adaptive behaviour wasn't observable in practice. All adaptive constants, helpers, and connection-quality listeners have been removed. The active room subscription now always requests 50 events.

### 3. Stop unsubscribing on navigation

`useSlidingSyncActiveRoom` previously called `unsubscribeFromRoom()` in its cleanup function, so every room visited would have its subscription removed even if the user briefly navigated away and came back. The hook now only switches the active-room subscription; subscriptions accumulate across the session so returning to a room is instant.

### 4. Auto-unsubscribe on leave / ban

`onMembershipLeave` handler registered on `RoomMemberEvent.Membership` in `attach()` / `dispose()`. When the local user leaves or is banned from a currently-subscribed room, `unsubscribeFromRoom()` fires automatically to clean up the sliding sync subscription.

## Additional Information

These changes are all loosely related to each other, and all work towards the goal of making slding sync more in-line with how it is handled in Element Web.

There are some things to note:

- Since we never remove rooms from the list, if the user joins a bunch of active rooms and then leaves the app open, it could use a decent amount of memory (50-100 MB), however, the SDK will cap timelines at 1000 entries, so it shouldn't be too bad.
- This significantly reduces the number of re-renders that were occuring, previously every time you entered a room sliding sync would need to add a room connection and then re-render. Now, since the room is connected and not unsubscribed, room content loads gracefully without a full re-render of the timeline.
- The Adaptive Timeline Limit has been removed, as it was not really noticeable in practice and added significantly more complexity to the design for little/no gain.
- If you leave a room, we kill the connection now, seems logical.

## Tests

6 unit tests in `slidingSync.test.ts`:

- `dispose()` calls `stop()` on the sliding sync instance
- Membership leave: unsubscribes on `leave`
- Membership leave: unsubscribes on `ban`
- Membership leave: ignores other-user membership changes
- Membership leave: ignores `join`
- Membership leave: no-op for rooms that were never subscribed


Just-Insane:
Did limited testing on the PR Preview, however, I'm not in that many large, active rooms, so my ability to test is limited. I will try leaving it open overnight to see what happens.

Overall, it seems more stable when switching between rooms, which is a major win imo.

dozro (approved):

Just-Insane:
After leaving the app open overnight (though, my rooms aren't overly active), it seems RAM usage is lower:

image

7w1 (approved):

PR #309, opened 2026-03-16
Implement per-message profile management and sending with a per-message profile

dozro:
### Description

Fixes https://github.com/SableClient/Sable/issues/298
Implements usage of [MSC4144](https://github.com/matrix-org/matrix-spec-proposals/pull/4144)

for future development: #340 and #341 

This pull request introduces support for per-message profiles (PMP), allowing users to set and use different profiles (such as display name, avatar, and pronouns) on a per-message basis. It adds both backend logic and UI components for managing these profiles, updates the message sending and rendering logic to support PMP, and exposes new settings and commands for users to interact with this feature. Additionally, it refactors some settings components for better modularity.

The most important changes are:

**Per-Message Profile (PMP) Support:**

* Added logic to send messages with a per-message profile if one is selected, including updating the message content to include the PMP data (`com.beeper.per_message_profile`).
* Added new commands (`/usepmp`, `/addpmp`, `/delpmp`) to manage and use per-message profiles via the chat input. 

**Settings UI Enhancements:**

* Introduced a new "Persona" section in the settings, with a dedicated page (`PerMessageProfilePage`) and overview/editor components for managing per-message profiles. (open to suggestion where that belongs lol)

**Pronoun Handling Refactor:**

* Moved `PronounSet` and `parsePronounsInput` to a shared utility file for consistent usage across the codebase, and improved pronoun filtering logic in the timeline. 

**Settings Refactoring:**

* Refactored the animal cosmetics ("cat" settings) into a standalone component for better modularity and maintainability. 

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #369, opened 2026-03-18
feat: add cosmetics support to per-message profiles

poolcritter:
This patch allows the following keys to be set on messages' `com.beeper.per_message_profiles` key:

* `moe.sable.app.name_color`: Overrides the name color; same format as `moe.sable.app.name_color` in profiles.
* `io.fsky.nyx.pronouns`: Overrides the per-message pronouns; same format as `io.fsky.nyx.pronouns` in profiles.

# Considerations

* The ability to override custom fonts has intentionally been left unimplemented. Per-message fonts could allow users to specify custom name fonts without needing room-admin approval, which could be dangerous.

### DIFF
--- a/src/app/features/room/message/Message.tsx
+++ b/src/app/features/room/message/Message.tsx
@@ -372,6 +372,8 @@ function MessageInternal(
             avatar_url: string | undefined;
             displayname: string | undefined;
             id: string | undefined;
+            "moe.sable.app.name_color": string | undefined;
+            "io.fsky.nyx.pronouns": any | undefined;
           }
         | undefined,
     [mEvent]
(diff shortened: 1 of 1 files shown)


dozro:
is that related/an alternative to https://github.com/SableClient/Sable/pull/309?

src/app/features/room/message/Message.tsx
```diff
@@ -430,8 +434,9 @@ function MessageInternal(
     return getParsedPronouns(rawName, parsePronouns);
   }, [pmp, senderDisplayName, parsePronouns]);
 
+  const sourcePronouns = pmp?.["io.fsky.nyx.pronouns"] ?? profile.pronouns ?? []
   const mergedPronouns = useMemo(() => {
-    const existing = profile.pronouns ? [...profile.pronouns] : [];
```
dozro, line 434: #309 already handles per-message-profile's pronouns iirc

dozro (changes requested):

src/app/features/room/message/Message.tsx
poolcritter: oh yeah, didn't notice — should I rebase onto there?

dozro: if you wish to :3

PR #394, opened 2026-03-19
refactor(timeline): replace useVirtualPaginator with virtua VList [TEST ONLY — NOT BEING PURSUED]

Just-Insane:
> **⚠️ NOTE: This is a purely exploratory/test PR. This refactor is not being actively pursued and is not intended to be merged. It exists only as a reference for what the migration would look like.**

---

## What this does

Replaces the custom `useVirtualPaginator` hook + CSS scroll container (`<Scroll>`) with [virtua](https://github.com/inokawa/virtua)'s `VList` component for the room message timeline.

## Key changes

- **`Timeline` type**: Removed `range: ItemRange` — virtua renders all loaded events, so the windowing range concept is gone entirely.
- **`VList` replaces `<Scroll><Box>`**: virtua's `VList` is both the scroll container and virtualizer. Items are rendered via a `data={indices}` + render-function children pattern.
- **Prepend anchoring**: Uses virtua's `shift` prop instead of CSS `overflow-anchor` (which doesn't work with virtua's absolute-positioning layout).
- **`atBottom` detection**: Replaced `IntersectionObserver` on a bottom anchor element with a pixel-distance check in `onScroll` (`scrollSize - offset - viewportSize < 100`).
- **Pagination triggers**: Backward/forward pagination is triggered from `onScroll` when the user scrolls within 500px of the top/bottom edge (replaces the old `IntersectionObserver` on front/back anchor elements).
- **Scroll-to-item**: Uses `VListHandle.scrollToIndex(index, { align, smooth })` instead of imperative DOM `scrollTop` manipulation.
- **Removed**: `scrollRef`, `atBottomAnchorRef`, `timelineItems`/`getItems()`, `useIntersectionObserver` calls, media `ResizeObserver`, `scrollToBottom` utility usage.

## Why this was explored

The existing `useVirtualPaginator` hook uses a manual sliding window (`range.start`/`range.end`) to limit how many DOM nodes are mounted. This works but has several edge cases around:
- Scroll position preservation when prepending events
- Spurious `IntersectionObserver` fires during bulk loads causing false-positive scroll resets
- Complex `useCallback` dependency chains that re-create observers on every events-length change

virtua handles all of this internally with a simpler API surface.

## Status

Draft / closed for reference only. Compiles cleanly (`pnpm tsc --noEmit` with node 24). Not tested at runtime.


Just-Insane:
For how quickly this was done, there are only a few major issues:
- [x] it sometimes **really** sticks to the bottom of the timeline
- [x] loading initial timeline (on first load), has the single item (from sliding sync's initial timeline limit of 1) near the top of the page
- [x] loading history (back pagination) seems to have a weird narrowing of the timeline view which then expands back to the correct layout once things have loaded

All three resolved in commit 45ace039.

Just-Insane:
Seems like most major issues with both classic sync and sliding sync are now worked out.

Still seeing similar issues in sliding sync as we see in `dev`, should be fixed by the inverted timeline work @7w1 is doing?

Just-Insane:
This was superseded by @7w1's timeline rewrite

PR #409, opened 2026-03-20
feat: sync settings across devices via Matrix account data

### LINKED ISSUE #402: Sync app settings across devices
### Describe the problem

Right now when you start using Sable on multiple devices or different instances, you have to configure each one manually and nothing is shared between them besides profile things like bio and such.

Add a versioned `im.sable.settings` key in account data to be able to persistent the settings across all devices running sable for a unified and seamless experience.

Note, we should avoid syncing things like telemetry or anything that has potential for abuse, since we don't have control over what other actors might do to `im.sable.settings`. Also have a good invalidation system in case the key data is invalid.

### Alternatives considered

_No response_

### Additional context

_No response_

Just-Insane:
Closes #402

## What

Adds opt-in settings synchronisation across devices using Matrix account data (`moe.sable.app.settings`), plus a JSON export/import fallback for users who prefer manual control.

## How it works

- **Sync toggle** (off by default) in General settings enables live sync
- On toggle-on, settings are loaded from account data immediately
- Local changes are debounced 2 s then uploaded to account data
- **Echo-token loop prevention**: each upload embeds a random token; when the homeserver echoes the event back the hook recognises it and skips re-applying (avoiding a feedback loop)
- Events arriving from other devices are applied and merge over local state
- **Non-syncable keys** are always taken from local state and never uploaded: `usePushNotifications`, `useInAppNotifications`, `useSystemNotifications`, `pageZoom`, `isPeopleDrawer`, `isWidgetDrawer`, `memberSortFilterIndex`, `developerTools`, `settingsSyncEnabled`

## JSON export / import

Available regardless of whether sync is enabled. Export triggers a browser download; import opens a file picker and merges the selected file into current settings (same schema validation as the sync path).

## Files changed

| File | Purpose |
|---|---|
| `src/types/matrix/accountData.ts` | `SableSettings = 'moe.sable.app.settings'` enum value |
| `src/app/state/settings.ts` | `settingsSyncEnabled: boolean` (default `false`) |
| `src/app/utils/settingsSync.ts` | serialize / deserialize / export / import helpers |
| `src/app/hooks/useSettingsSync.ts` | side-effect hook + status atoms |
| `src/app/pages/client/ClientNonUIFeatures.tsx` | wire `SettingsSyncFeature` for session lifetime |
| `src/app/features/settings/general/General.tsx` | `SettingsSyncSection` UI |
| `src/app/utils/settingsSync.test.ts` | 26 tests for pure utility functions |
| `src/app/hooks/useSettingsSync.test.tsx` | 13 tests for hook behaviour (mount load, debounce, echo-token, remote updates) |

#### Type of change

#### AI assistance level

- [ ] **None** — No AI tools were used
- [ ] **Research only** — AI used for research, brainstorming, or understanding existing code; no generated code included
- [ ] **AI-assisted** — AI suggested code or logic that I reviewed, tested, and understand fully; I can explain every change
- [X] **Primarily AI-generated** — The majority of code was AI-generated; I have thoroughly reviewed it, verified correctness, and made it my own

### Checklist:

### DIFF
--- a/src/app/hooks/useSettingsSync.ts
+++ b/src/app/hooks/useSettingsSync.ts
@@ -1,0 +1,105 @@
+import { useCallback, useEffect, useRef } from 'react';
+import { atom, useAtom, useSetAtom } from 'jotai';
+import { MatrixEvent } from '$types/matrix-sdk';
+import { AccountDataEvent } from '$types/matrix/accountData';
+import { useMatrixClient } from '$hooks/useMatrixClient';
+import { useAccountDataCallback } from '$hooks/useAccountDataCallback';
+import { settingsAtom } from '$state/settings';
+import { deserializeFromSync, serializeForSync } from '$utils/settingsSync';
+
+export type SyncStatus = 'idle' | 'syncing' | 'error';
+
+/** Milliseconds to wait after a local settings change before uploading. */
+const DEBOUNCE_MS = 2000;
+
+/** Unix timestamp (ms) of the last confirmed sync, or null if never synced this session. */
+export const settingsSyncLastSyncedAtom = atom<number | null>(null);
+
+/** Current upload state for UI feedback. */
+export const settingsSyncStatusAtom = atom<SyncStatus>('idle');
+
+/**
+ * Side-effect hook that:
+ *  - loads settings from account data when sync is first enabled
+ *  - listens for live updates arriving from other devices
+ *  - debounce-uploads local changes back to account data
+ *
+ * Only active when `settings.settingsSyncEnabled === true`.
+ * Call this once from a component that stays mounted for the session lifetime.
+ */
+export function useSettingsSyncEffect(): void {
+  const mx = useMatrixClient();
+  const [settings, setSettings] = useAtom(settingsAtom);
+  const setLastSynced = useSetAtom(settingsSyncLastSyncedAtom);
+  const setSyncStatus = useSetAtom(settingsSyncStatusAtom);
+
+  // Keep a ref so callbacks can always read the latest value without stale closures.
+  const settingsRef = useRef(settings);
+  settingsRef.current = settings;
+
+  const syncEnabled = settings.settingsSyncEnabled;
+
+  // ── On mount / when sync is first enabled: load from account data ──────────
+  useEffect(() => {
+    if (!syncEnabled) return;
+    const event = mx.getAccountData(AccountDataEvent.SableSettings as never);
+    if (!event) return;
+    const merged = deserializeFromSync(event.getContent(), settingsRef.current);
+    if (merged) {
+      setSettings(merged);
+      setLastSynced(Date.now());
+    }
+  }, [mx, syncEnabled, setSettings, setLastSynced]);
+
+  // ── Echo-detection: track the token of our last upload ────────────────────
+  // When our upload echoes back via ClientEvent.AccountData we skip applying it
+  // (to avoid overwriting settings that changed between upload and echo).
+  const pendingEchoTokenRef = useRef<string | null>(null);
+
+  // ── Live updates from other devices ───────────────────────────────────────
+  const onAccountData = useCallback(
+    (event: MatrixEvent) => {
+      if (event.getType() !== AccountDataEvent.SableSettings) return;
+      if (!settingsRef.current.settingsSyncEnabled) return;
+
+      const content = event.getContent();
+
+      // If this is the echo of our own upload, just confirm success and skip.
+      if (typeof content._echo === 'string' && content._echo === pendingEchoTokenRef.current) {
+        pendingEchoTokenRef.current = null;
+        setLastSynced(Date.now());
+        setSyncStatus('idle');
+        return;
+      }
+
+      // Otherwise it came from another device — apply it.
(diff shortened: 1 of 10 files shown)


Just-Insane:
Tests well, can change settings and see them syncing to account data, uploading and downloading the files also works and overwrites settings as expected.

An example settings export file is below:
[sable-settings-1773977439924.json](https://github.com/user-attachments/files/26131917/sable-settings-1773977439924.json)

Just-Insane:
This is feature complete, but I can't support it going forward. You can merge it if you think it'll be appreciated and maintainable.

src/app/hooks/useSettingsSync.ts
```diff
@@ -0,0 +1,105 @@
...
+  // Keep a ref so callbacks can always read the latest value without stale closures.
+  const settingsRef = useRef(settings);
+  settingsRef.current = settings;
+
+  const syncEnabled = settings.settingsSyncEnabled;
+
+  // ── On mount / when sync is first enabled: load from account data ──────────
+  useEffect(() => {
+    if (!syncEnabled) return;
+    const event = mx.getAccountData(AccountDataEvent.SableSettings as never);
```
dozro, line 45: The `as never` really seems off. 

From the typescript documentation, and the reason why it feels off:

> The never type represents the type of values that never occur. For instance, never is the return type for a function expression or an arrow function expression that always throws an exception or one that never returns. Variables also acquire the type never when narrowed by any type guards that can never be true.

src/app/hooks/useSettingsSync.ts
```diff
@@ -0,0 +1,105 @@
...
+  useEffect(() => {
+    if (!syncEnabled) return undefined;
+
+    clearTimeout(timerRef.current);
+    timerRef.current = setTimeout(() => {
+      setSyncStatus('syncing');
+      const token = Math.random().toString(36).slice(2, 10);
+      pendingEchoTokenRef.current = token;
+      const content = { ...serializeForSync(settingsRef.current), _echo: token };
+      mx.setAccountData(AccountDataEvent.SableSettings as never, content as never).catch(() => {
```
dozro, line 97: is there a reason for the `as never` here

`as never` implies that the type never occurs 

From the typescript documentation:

> The never type represents the type of values that never occur. For instance, never is the return type for a function expression or an arrow function expression that always throws an exception or one that never returns. Variables also acquire the type never when narrowed by any type guards that can never be true.

Is that intended here (genuine question)?

dozro (changes requested):

Just-Insane:
Two patches which fix @dozro's requested changes:

Makes the SDK aware of the additional account data
```
diff --git a/src/types/matrix-sdk-events.d.ts b/src/types/matrix-sdk-events.d.ts
index 532170e0..7b73052f 100644
--- a/src/types/matrix-sdk-events.d.ts
+++ b/src/types/matrix-sdk-events.d.ts
@@ -14,5 +14,6 @@ declare module 'matrix-js-sdk/lib/@types/event' {
     'im.ponies.user_emotes': PackContent;
     'im.ponies.emote_rooms': EmoteRoomsContent;
     'moe.sable.app.nicknames': Record<string, string>;
+    'moe.sable.app.settings': Record<string, unknown>;
   }
 }
```

Removes the `as never` - no longer needed once the SDK is aware of the new account data.
```
diff --git a/src/app/hooks/useSettingsSync.ts b/src/app/hooks/useSettingsSync.ts
index 925f6cb7..2e35ff23 100644
--- a/src/app/hooks/useSettingsSync.ts
+++ b/src/app/hooks/useSettingsSync.ts
@@ -42,7 +42,7 @@ export function useSettingsSyncEffect(): void {
   // ── On mount / when sync is first enabled: load from account data ──────────
   useEffect(() => {
     if (!syncEnabled) return;
-    const event = mx.getAccountData(AccountDataEvent.SableSettings as never);
+    const event = mx.getAccountData(AccountDataEvent.SableSettings);
     if (!event) return;
     const merged = deserializeFromSync(event.getContent(), settingsRef.current);
     if (merged) {
@@ -94,7 +94,7 @@ export function useSettingsSyncEffect(): void {
       const token = Math.random().toString(36).slice(2, 10);
       pendingEchoTokenRef.current = token;
       const content = { ...serializeForSync(settingsRef.current), _echo: token };
-      mx.setAccountData(AccountDataEvent.SableSettings as never, content as never).catch(() => {
+      mx.setAccountData(AccountDataEvent.SableSettings, content as Record<string, unknown>).catch(() => {
         pendingEchoTokenRef.current = null;
         setSyncStatus('error');
       });
```

AI Disclosure: Written by AI, though the changes are logical looking at the way other account data is configured (nicknames)

src/app/hooks/useSettingsSync.ts
Just-Insane: Resolved via https://github.com/SableClient/Sable/pull/409#issuecomment-4103767837

src/app/hooks/useSettingsSync.ts
Just-Insane: Resolved via https://github.com/SableClient/Sable/pull/409#issuecomment-4103767837

PR #462, opened 2026-03-21
Feat: Pin event styling and coloring events in timeline

nushea:
### Description
This PR adds formatting for the pin events so that you can see the events that have been added or removed at a glance (up to 4 as this is what i thought was a reasonable limit maybe)
image

It also adds the chosen colors of users to the events of them leaving, joining, changing their name, the room name et cetera
image

Fixes #

#### Type of change

### Checklist:


7w1:
Imo it seems a little odd to have two pin icons for every pin event? It just feels a little cluttered.

nushea:
> Imo it seems a little odd to have two pin icons for every pin event? It just feels a little cluttered.

I would love to find a better alternative but this is what compromise we could best think of in https://matrix.to/#/!a6sXbRuOyyc7MKutmy:sable.moe/$3U5-kgbS-OAeX8dvDCPKxcd936wnFwi-DkdiF_v5kVs?via=matrix.org&via=unredacted.org&via=4d2.org

maybe it could be made an animation or sth? but then it would still have the issue after that is done, so idk what could be put instead

Rosy-iso (commented):
UI wise, the way the pin idea is implemented looks good. Though the 📌+ and 📌- icons seem a little...redundant? the message above already says if it's being pinned or unpinned (and pin events arent something that happens often) and there's *another* pin icon just to the left already...I would personally remove the 📌+ and 📌- but I'm also not full on against it, open to discussion there.

For the colored names, I think it's a nice change that adds some color to the UI which is certainly welcomed...but why is it alongside the pin event? These two things have very little correlation... It's fine here since it's highly likely that both things could be merged but I'd personally keep PRs precise to a specific scope (though, this is outside my area of jurisdiction, its just my opinion)

nushea:
> UI wise, the way the pin idea is implemented looks good. Though the 📌+ and 📌- icons seem a little...redundant? the message above already says if it's being pinned or unpinned (and pin events arent something that happens often) and there's _another_ pin icon just to the left already...I would personally remove the 📌+ and 📌- but I'm also not full on against it, open to discussion there.

the reason why they are there is because you are allowed to have multiple pins in one event, so youd need to tell which message got added and which got removed, and just having the +/- would make it look even more out of place so thats why i have ended up on that but if there are any other ideas on how i could make it obvious the list is related to the parent event and show on each element whether it has been added or removed id love to find a better option

> For the colored names, I think it's a nice change that adds some color to the UI which is certainly welcomed...but why is it alongside the pin event? These two things have very little correlation... It's fine here since it's highly likely that both things could be merged but I'd personally keep PRs precise to a specific scope (though, this is outside my area of jurisdiction, its just my opinion)

true, i put them into different commits but i shouldve put them into different prs and like honestly the only reason they arent 2 different prs is that i made both changes in the same period of time, but from a logical standpoint they should be 2, and i should try to restrict the scope of my prs whenever i send them to be per feat/fix

Rosy-iso:
> the reason why they are there is because you are allowed to have multiple pins in one event, so youd need to tell which message got added and which got removed, and just having the +/- would make it look even more out of place so thats why i have ended up on that but if there are any other ideas on how i could make it obvious the list is related to the parent event and show on each element whether it has been added or removed id love to find a better option

I seee, that makes a lot more sense now yeah, that works!

Rosy-iso (approved):

7w1 (approved):

PR #468, opened 2026-03-21
Fix: fix file messages to show better on other clients

nushea:
### Description
This change fixes the fact that on some clients (fluffychat) if you have a formatted body at all no matter what it is it will display it, it just removes the fields until they are used.

#### Type of change

### Checklist:

### DIFF
--- a/.changeset/small-image-fix.md
+++ b/.changeset/small-image-fix.md
@@ -1,4 +1,5 @@
 ---
 default: patch
 ---
+
 Fix messages sent from sable showing wrong on other client(s)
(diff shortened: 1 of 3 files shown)


dozro (approved):
doesn’t add much value anyway if we just write in `formatted_body` the same thing as in `body` without any formatting lol

it’s in both times just `file.name`

nushea:
> doesn’t add much value anyway if we just write in `formatted_body` the same thing as in `body` without any formatting lol
> 
> it’s in both times just `file.name`

true that it was redundant

7w1 (approved):
fixing bugs in other clients smh

PR #473, opened 2026-03-22
Add maskable icons for better iOS/Android PWA appearance

Septicity:
### Description

I've added maskable icons to the selection of app icons. Maskable icons are designed to be cropped into any shape, like on various Android launchers, and this distinction additionally makes them work well on iOS.

Here is a screenshot for comparison, with the new icon on the left :
image

In the process of doing this, I've adjusted the structure of the directories containing the app icons (they are no longer sorted by Android/Apple, since any platform can pull from either format under different circumstances). On that note, it looks like many of the resolutions could probably be pruned... I kept all of the existing ones, though.

I've also removed the dots from the paths in the HTML file, since this was causing major issues in the `dev` run. I haven't been able to do a proper build and test by hosting the /dist folder, so if this does break anything in that scenario, I can revert that change.

I've tested this to completion on Linux and iOS so far, and I will edit this part as I test more platforms (assuming it isn't merged by then).
- [ ] Android: Seems to work; a bit stretched, though? Edge might just be bad at making PWAs. Should try more browsers.
- [ ] Windows:
- [ ] macOS:
- [ ] iPadOS:

#### Type of change

### Checklist:


index.html
```diff
@@ -18,15 +18,15 @@
 
     <meta property="og:title" content="Sable" />
     <meta property="og:url" content="https://app.sable.moe" />
-    <meta property="og:image" content="./public/favicon.png" />
```
dozro, line 21: maybe don’t turn the relative paths to absolute paths here. we want to be able to theoretically to host it at `example.org/sable` :3

same for the other path changes in this file :3

dozro (changes requested):

index.html
Septicity: If these paths get resolved in an actual build, then fair. I was just having trouble getting it to link files correctly in the temporary `dev` builds (it wouldn't load the favicon or apple-touch-icon images whatsoever, except for somehow managing to in the body of the page).

I should probably also change the paths in the `manifest.json` to what they would be after Vite's config is applied, as they were before.

dozro:
with the `manifest.json` we need to pay attention not to have a regression so this happens 34b2901566bb49bdeab4cd6d3fb9dc7ee845279f

Septicity:
The regression shouldn't happen; my changes haven't touched how `manifest.json` is accessed or stored. I specifically avoided putting a dot on its path since it seemed to work correctly already.

dozro (approved):

7w1 (approved):

Issue #391, opened 2026-03-19
Add a Setup Wizard


hazre:
### Describe the problem

We have lots of options in Sable and most people won't know about them without digging through through the settings.

### Describe the solution you'd like

I think we should expose some of the big impacting options, when a user first starts using Sable, to customize their experience. It's also somewhat of an advertisement for what sable offers.

### Alternatives considered

Suffer.

wolterkam:
Settings and customization aren't very sane at the moment and things are constantly in motion. Users should preferably stumble upon the unique features organically, imo.

PR #474, opened 2026-03-22
Sidebar settings and features discoverability

wolterkam:
### Description
Improves discoverability of settings and features related to Sidebar. Somewhat related to #391.

image

Fixes #

#### Type of change

### Checklist:

### DIFF
--- a/.changeset/feat-sidebar-settings-discoverability.md
+++ b/.changeset/feat-sidebar-settings-discoverability.md
@@ -1,0 +1,5 @@
+---
+'default': minor
+---
+
+Add sidebar three dot menu for quick access to related settings
--- a/.changeset/feat-sidebar-settings-discoverability.md
+++ b/.changeset/feat-sidebar-settings-discoverability.md
@@ -1,5 +1,5 @@
 ---
-'default': minor
+default: minor
 ---
 
 Add sidebar three dot menu for quick access to related settings
--- a/src/app/pages/client/SidebarNav.tsx
+++ b/src/app/pages/client/SidebarNav.tsx
@@ -28,6 +28,8 @@ export function SidebarNav() {
   const [showPingCounts, setShowPingCounts] = useSetting(settingsAtom, 'showPingCounts');
 
   const handleContextMenu: MouseEventHandler<HTMLDivElement> = (evt) => {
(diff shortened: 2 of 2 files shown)


dozro:
@Rosy-iso this seems like a ui change, so yours to review :3

Rosy-iso (changes requested):
Hm, this one is a tough one.
The idea in itself is a nice one, can help newer users...but

First of all, I don't think I've seen other popup menus have toggles before...but there's always a first time and I think it's been implemented fine here (using the bigger switch would be too cumbersome anyway)

But the button to access this menu...that i'm really unsure about...
First of all the three dots are rotated 90° compared to all other thee dots buttons (unless I missed any), which I feel like should be changed for Consistency, but other than that...it's complicated: its a tiny borderless button which is consistent with all other three dots button but is inconsistent with the other buttons in the sidebar..but that's fine since this is a smaller menu with less of a scope, it fits to be smaller...but it still occupies a decent amount of space for something that most of the time you do only need to click once and barely ever again.

To summarize my ramblings, maybe the button needs a bit more purpose to justify its presence? It does seem to essentially take up an entire "slot" so it is substantial imo, but I'm willing to discuss all this

Overall, still def a good and promising idea I believe

wolterkam:
@Rosy-iso Hm, I will remove this button for now and leave the options as a right click context menu. If some more features come up that could be put in there, we could then go back to the button.

hazre:
Maybe it can be context aware? Like right clicking on the sidebar but then again how would you handle it on mobile

wolterkam:
> Maybe it can be context aware? Like right clicking on the sidebar but then again how would you handle it on mobile

I assume long press? Does the app actually translate the touch-and-hold events to right-click at the moment, or are all context menus non functional on mobile?

7w1:
> I assume long press? Does the app actually translate the touch-and-hold events to right-click at the moment, or are all context menus non functional on mobile?

Nothing is translated, context menus are non-functional unless they have a specific mobile implementation

wolterkam:
I have moved these quick settings to live in a context menu activated with a right mouse click.

7w1 (approved):

PR #475, opened 2026-03-22
Fix/sw media auth - issues causing unauthenticated media requests

### LINKED ISSUE #358: Some media fails to load (Mismatched SHA-256 digest)
### Describe the bug

Occasionally, some media will fail to load, namely user avatars and photos in chats. It doesn't seem to impact all media in a given chat.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
- OS: MacIntel
- Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/144.0.0.0 Safari/537.36
- Sable: v1.9.3 (e086398)
```

### Additional context

Error:

Uncaught (in promise) Error: Mismatched SHA-256 digest
    at index-DfcHwRe0.js:77814:46
    at async Gp (index-DfcHwRe0.js:77902:15)
    at async Yp (index-DfcHwRe0.js:77980:12)
    at async index-DfcHwRe0.js:109521:19
    at async index-DfcHwRe0.js:36749:13

Just-Insane:
### Description

Appears to fix #358

Main issues:

1. There is an early-return bug that occurred when the requested URL didn't match the session's baseURL which caused an unauthenticated request
2. Resolves issues arising from using a stale or missing SW session
3. Populate a session from cache on restarts to prevent a timing issue
4. An issue with `clientId` being empty from clients such as prerenders which caused unauthenticated media requests

Note that you need to reload the SW by either killing the current one in DevTools, or if on mobile, the simplest option is to remove/re-download the PWA.

#### Type of change

### Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas

### AI disclosure:

These changes (2 commits) combine to reduce the likelihood that a missing/stale session will result in failed authenticated media requests that result in 401 errors by:

- requesting sessions from all window clients so that the sessions map is populated after an SW restart
- persisting the SW session to cache, which is then tried when requesting media (mainly if still waiting for sessions from the client window) - the `preloadedSession` in-memory fallback is discarded once a live session is received, though the underlying cache entry remains until logout.
- tries to use a persisted session from the cache on SW activation to help resolve a timing issue between SW restart and the first setSession return
- Note that if the persisted session is no longer valid, authenticated media requests will still fail with the same 401 unauthenticated response.
- fixes logic bugs that 1: caused the browser to make unauthenticated requests if the URL didn't match the cached session's baseUrl, and 2: attempts to use the preloaded session rather than allowing the browser to send unauthenticated requests when the clientId is empty (i.e. prerenders).

### DIFF
--- a/.changeset/fix-sw-media-auth-401.md
+++ b/.changeset/fix-sw-media-auth-401.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix service worker authenticated media requests returning 401 errors after SW restart or when session data is missing/stale.
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -460,6 +460,11 @@ self.addEventListener('activate', (event: ExtendableEvent) => {
     (async () => {
       await self.clients.claim();
       await cleanupDeadClients();
+      // Proactively request sessions from all window clients so the sessions Map
+      // is pre-populated after a SW restart, rather than waiting for the first
+      // media fetch to trigger requestSessionWithTimeout.
+      const windowClients = await self.clients.matchAll({ type: 'window' });
+      windowClients.forEach((client) => client.postMessage({ type: 'requestSession' }));
     })()
   );
 });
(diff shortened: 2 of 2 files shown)


Just-Insane:
This seems to prevent background notifications from working as expected.

PR #476, opened 2026-03-22
feat: add editor multiline mode and many edge case fixes

### LINKED ISSUE #350: voice message recording not possible on phone bc ui design
### Describe the bug

if i tap one time on the recording button, it will immediately dismiss. if i hold it will try to select text…. so not really usable.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
on iPad, using the whatever the current version at app.sable.moe is
```

### Additional context

_No response_

hazre:
### Description

Fixes #350

This adds a dynamic multiline layout to `CustomEditor`. When the message input grows past a single line (word wrap, explicit newlines, or multiple blocks), the editor shifts from a single-row layout into a two-row grid, before/textarea on top, with responsive content spanning the full width below. It drops back to single-row once the text fits on one line again.

The old `replacementContent` prop (which swapped out the textarea on mobile during voice recording) is gone. It's replaced by `responsiveAfter`, which renders inline next to the send button in single-line mode and moves down to the footer row in multiline mode. There's also a `forceMultilineLayout` prop for when you want to push into multiline immediately, which is what we use when the audio recorder opens. This means the recorder now takes the same code path on desktop and mobile instead of having a separate mobile-only branch.

A few other things got fixed along the way:

- Holding the record button on mobile was selecting the "following the conversation" text underneath it. Fixed with `userSelect: none`. (related to #350)
- The stop button's `onClick` was being skipped on mobile entirely, so if `pointerup` got lost mid-drag there was no way to stop recording. Now `onClick` works as a fallback when recording is already active. (a form of recovery in case more situations like #350 happen again)
- The waveform bar count is now calculated from the actual container width via `ResizeObserver` instead of a hardcoded number.
- `cleanupMediaRecorder` wasn't being called in all the stop/delete/restart/error/unmount paths, which could leave the mic track open. That's fixed.
- Firefox couldn't seek voice messages because `audio/webm` doesn't include a seek index. Codec preference now puts `audio/ogg;codecs=opus` first so Firefox records seekable blobs. Also added a `seekTo` helper that waits for `loadedmetadata` if the audio element isn't ready yet, and `audio.preload = 'auto'` + `audio.load()` on mount so Firefox parses metadata right away.
- Multiline wrap detection is now debounced through `requestAnimationFrame` and uses a `ResizeObserver` on the scroll container to cut down on flicker and extra re-renders.
- Added cloudflared tunnels for easier testing on mobile

Also sorry for the cursed line height measurer, there was legit no other way to detect it that didn't cause other issues.

Good amount of this was created with help of AI (like figuring out options on how to fix the edge cases), especially the edge case unit tests, but It had to be fixed by hand in many places and frankly I can't tell you exactly where was what. I've been working on this for couple days now.. (losing my mind over it)

### Screenshots:
https://github.com/user-attachments/assets/bcff3766-6c60-46c7-ac9e-a8b943a665d3

https://github.com/user-attachments/assets/367cdcc4-666e-4324-8e81-1b415feb0aca

#### Type of change

### Checklist:

### AI disclosure:


dozro:
# functional testing

functions well:
- multiline editor
- waveform

couldn't be tested rn:
- mobile phone specific bugs regarding voice messages
- firefox specific bugs

dozro (approved):
code lgtm. 

could use a bit more of code documention :3

dozro (approved):

7w1 (approved):

PR #485, opened 2026-03-23
Fix Video Camera auto enabling on Electron (and Tauri) clients using Sable Call.

### LINKED ISSUE #285: Camera is on by default 💀
When you join any call, your camera gets always turned on by **default**.
This is a behavior Sable shares with Element and I think its crazy 

Even when the UI clearly shows you that the cam is off,
Image
and you then join, it still gets turned on.

you are directly calling somebody?
The Camera gets automatically turned on.

This **shouldn't** be the default behaviour

Rawrington:
### Description

Fixes #285 for good. Element Call (and thus Sable Call) use a `callIntent` paramter derived from the `intent` parameter to decide if the Camera should be requested. The Camera was being requested on every call, and thus auto approved on Electron clients (and Tauri clients). This affected both starting a room call and joining a VC room. This is likely  due to the fact that in Element they are 'Video Rooms'. This change simply uses `JoinExistingDMVoice` which sets the correct parameters to join a VC room OR call from the room header. The `StartCall` and `JoinExisting` intents function the same on Sable Call's end, this can be changed if they start to diverge in future updates of course though.

I have tested this change in both scenarios of calling with multiple people and it has functioned as intended.

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/.changeset/fix-video-auto-enabled.md
+++ b/.changeset/fix-video-auto-enabled.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix Camera being enabled by default even when the client has it off pre joining in browsers that permit the video (Electron/Tauri as examples).
--- a/src/app/hooks/useCallEmbed.ts
+++ b/src/app/hooks/useCallEmbed.ts
@@ -52,7 +52,7 @@ export const createCallEmbed = (
   const ongoing =
     MatrixRTCSession.sessionMembershipsForRoom(room, rtcSession.sessionDescription).length > 0;
 
-  const intent = CallEmbed.getIntent(dm, ongoing);
+  const intent = CallEmbed.getIntent(dm, ongoing, pref.video);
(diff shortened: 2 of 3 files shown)


henk717:
Works for me on tauri where most solutions fail.
Unrelated to this i'd like to see that change that was discussed on discord in the past where it won't remember the camera state when its not obvious. For example joining from a room list you don't see the current camera button state and can end up joining with a camera enabled without noticing.

Rawrington:
> Works for me on tauri where most solutions fail. Unrelated to this i'd like to see that change that was discussed on discord in the past where it won't remember the camera state when its not obvious. For example joining from a room list you don't see the current camera button state and can end up joining with a camera enabled without noticing.

that would definitely be a seperate solution. Since with this change Sable Call respects the video on/off of the user on joining. It could be worth always forcing video off after leaving a call so it's more inline with discord but I didn't feel that was in scope of a small change like this

7w1 (approved):

PR #487, opened 2026-03-23
Fix the autocomplete "dropdown" menus

nushea:
### Description
This PR seeks to fix the issues around the autocomplete for emoji, room, command, and people. 

The issues fixed by this PR are: the autocomplete menu not scrolling, not rollovering (ie when you were at the top you couldnt go back to the bottom or vice versa), and when editing a message and trying to select an emoji the arrow buttons not working.

#### Type of change

### Checklist:

### AI disclosure:

Written with all the natural stupidity possible in one so-called human


7w1 (commented):
Haven't tested yet but looks good on a glance

7w1:
Seems like enter can't select an emoji when editing a message, don't know if thats intended.
When the autocomplete menu pops up, there isn't any selection box around an entry until you start using arrow keys, i feel like having that selection box appear around the first element by default would make things a little more clear?

nushea:
> Seems like enter can't select an emoji when editing a message, don't know if thats intended. When the autocomplete menu pops up, there isn't any selection box around an entry until you start using arrow keys, i feel like having that selection box appear around the first element by default would make things a little more clear?

if the objective is to use an actual focustrap as it is intended, you cant both have it not focused and selecting the first item as far as i can tell, the previous pr only had the 1st item selected by hard-selecting it and then not using the focus trap in any meaningful way, i could try to rewrite this to avoid using a focustrap at all, but any pr that wants to implement the changes that the previous pr wanted to implement without it being a complete workaround of the focustrap would require just not using one and rewriting the autocomplete function (again as far as i can tell and my knowledge around focustraps goes), which i would be happy to try my hand at if highlighting the 1st item is a hard requirement (also i will be looking into why entering doesnt select an item on edit messages)

nushea:
I will make a new pr whenever i have the time to do it properly since this would still need quite some rewriting, but this fixes all the issues as far as i can tell (except not highlighting the 1st item as you type it since that is still not how focustraps work at all afaik but could use a rewrite to avoid using a focustrap for this))

7w1 (approved):

PR #514, opened 2026-03-24
feat: room abbreviations with hover tooltips and space inheritance

Just-Insane:
### Description

Adds a room abbreviations feature: moderators can define term/definition pairs in room settings. Defined terms are highlighted with a dotted underline in messages and show a tooltip on hover (tap-to-pin on mobile). Abbreviations defined in a parent space are inherited by child rooms and nested spaces, with overrides allowed at each level. The settings UI shows a flat list of inherited and locally-defined entries.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

The code adds changes that allow for setting and viewing room abbreviations, with an on-hover tooltip (or press on mobile - semi-persistent). It stores these in the room state under a new key, and recursively fetches abbreviations from the space level. This also adds regex so that abbreviations correctly work at word boundaries. Also adds the needed hooks to `room.ts` and `slidingSync.ts`

### DIFF
--- a/src/app/components/message/RenderBody.tsx
+++ b/src/app/components/message/RenderBody.tsx
@@ -41,7 +41,11 @@ function AbbreviationTerm({ text, definition }: AbbreviationTermProps) {
       <TooltipProvider position="Top" tooltip={tooltipContent}>
         {(triggerRef) => (
           // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
-          <abbr ref={triggerRef as React.Ref<HTMLElement>} onClick={handleClick}>
+          <abbr
+            ref={triggerRef as React.Ref<HTMLElement>}
+            onClick={handleClick}
+            style={{ textDecoration: 'underline dotted', cursor: 'help' }}
+          >
             {text}
           </abbr>
         )}
--- a/src/app/components/message/RenderBody.tsx
+++ b/src/app/components/message/RenderBody.tsx
@@ -1,4 +1,4 @@
-import { MouseEventHandler, useEffect, useState } from 'react';
+import { MouseEventHandler, ReactNode, useEffect, useState } from 'react';
 import parse, { HTMLReactParserOptions } from 'html-react-parser';
 import Linkify from 'linkify-react';
 import { Opts } from 'linkifyjs';
@@ -57,6 +57,33 @@ function AbbreviationTerm({ text, definition }: AbbreviationTermProps) {
       )}
     </>
   );
+}
+
+/**
+ * Builds a `replaceTextNode` callback for use with {@link getReactCustomHtmlParser}.
+ * Returns `undefined` when there are no abbreviations to apply (avoids creating
+ * extra closures in the common case).
+ */
+export function buildAbbrReplaceTextNode(
+  abbrMap: Map<string, string>
+): ((text: string) => ReactNode | undefined) | undefined {
+  if (abbrMap.size === 0) return undefined;
+  return (text: string) => {
+    const segments = splitByAbbreviations(text, abbrMap);
+    if (!segments.some((s) => s.termKey !== undefined)) return undefined;
+    return (
+      <>
+        {segments.map((seg, i) =>
+          seg.termKey !== undefined ? (
+            // eslint-disable-next-line react/no-array-index-key
+            <AbbreviationTerm key={i} text={seg.text} definition={abbrMap.get(seg.termKey) ?? ''} />
+          ) : (
+            seg.text
+          )
+        )}
+      </>
+    );
+  };
 }
 
 type RenderBodyProps = {
--- a/src/app/components/message/RenderBody.tsx
+++ b/src/app/components/message/RenderBody.tsx
@@ -1,8 +1,11 @@
 import parse, { HTMLReactParserOptions } from 'html-react-parser';
 import Linkify from 'linkify-react';
 import { Opts } from 'linkifyjs';
+import { Text, Tooltip, TooltipProvider, toRem } from 'folds';
 import { sanitizeCustomHtml } from '$utils/sanitize';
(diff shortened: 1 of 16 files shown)


dozro:
requested review from @Rosy-iso because it seems to be an ui change and from @TastelessVoid because they (according to github) recently worked on the changed files

Rosy-iso:
I'll give it a look soon

Rosy-iso (changes requested):
Interesting feature! I like the concept a lot actually, is this idea from you or is it from some matrix spec?

That aside, consistency wise things are mostly okay: the abbreviations list is perfectly consistent while the abbreviation creation modal is...almost consistent. While the fact that the modal itself is not split into multiple sections might seem the issue at first, that is actually perfectly fine as this *is* something with precedence that is rarely done for the few settings that *must* work with multiple elements at once (room settings, emoji pack settings)

The inaccuracy is in the spacing and placement: the elements of the modal are too close together and the button should be on the left, not the right, use the other modals as reference

Just-Insane:
The abbreviation concept is mine, I wanted a lightweight way for rooms to define their own jargon/acronyms without needing a full wiki or pinned message. Saw a few people asking about PMP messages were while that was being developed.

Fixed both spacing issues: bumped the form's internal gap from `200` to `400` (matching the pattern in e.g. the room profile form), and removed `justifyContent="End"` from the button row so it sits on the left like the other modals.

Just-Insane:
Seems to be an issue with abbreviations not showing properly, trying to track that down.

Edit: looks good now, looks like it was an issue with messages sent via PMPs

Rosy-iso (approved):
Seems to be consistent now indeed, this works :)

(This approval is purely UI wise, the code should still be audited and reviewed separately by someone else!)

src/app/components/message/RenderBody.tsx
```diff
@@ -1,10 +1,91 @@
...
+): ((text: string) => JSX.Element | undefined) | undefined {
+  if (abbrMap.size === 0) return undefined;
+  return function replaceTextNode(text: string) {
+    const segments = splitByAbbreviations(text, abbrMap);
+    if (!segments.some((s) => s.termKey !== undefined)) return undefined;
+    return (
+      <>
+        {segments.map((seg, i) =>
+          seg.termKey !== undefined ? (
+            // eslint-disable-next-line react/no-array-index-key
```
7w1: I feel like we can probably avoid these `// eslint-disable-next-line react/no-array-index-key`s? the splitByAbbreviations function can generate unique ids using indexes like `txt-{index}` and pass them through so we aren't using just plain indexes in the actual react component constructors

src/app/components/message/RenderBody.tsx
```diff
@@ -20,12 +101,41 @@ export function RenderBody({
...
+  if (abbrMap.size > 0) {
+    const segments = splitByAbbreviations(body, abbrMap);
+    if (segments.some((s) => s.termKey !== undefined)) {
+      return (
+        <>
+          {segments.map((seg, i) => {
+            if (seg.termKey !== undefined) {
+              const definition = abbrMap.get(seg.termKey) ?? '';
+              return (
+                // eslint-disable-next-line react/no-array-index-key
```
7w1: remove this

src/app/components/message/RenderBody.tsx
```diff
@@ -20,12 +101,41 @@ export function RenderBody({
...
+                <AbbreviationTerm key={i} text={seg.text} definition={definition} />
+              );
+            }
+            return (
+              // eslint-disable-next-line react/no-array-index-key
```
7w1: remove this too

src/app/features/room-settings/abbreviations/RoomAbbreviations.tsx
```diff
@@ -0,0 +1,328 @@
...
+                                </Box>
+                              )}
+                            </SequenceCard>
+                          ))}
+                          {!isSpace &&
+                            ancestorGroups.flatMap(({ spaceId, spaceName, entries: spEntries }) =>
+                              spEntries.map((entry, index) => (
+                                <SequenceCard
+                                  // eslint-disable-next-line react/no-array-index-key
+                                  key={`${spaceId}-${index}`}
```
7w1: Same here, we can use entry.term instead of the index

src/app/features/room-settings/abbreviations/RoomAbbreviations.tsx
```diff
@@ -0,0 +1,328 @@
...
+                          <Text size="T300" style={{ color: 'var(--mx-surface-variant-on)' }}>
+                            No {isSpace ? 'space' : 'room'}-level abbreviations defined yet.
+                            {canEdit && ' Use the form above to add some.'}
+                          </Text>
+                        </SequenceCard>
+                      ) : (
+                        <>
+                          {entries.map((entry, index) => (
+                            <SequenceCard
+                              // eslint-disable-next-line react/no-array-index-key
```
7w1: Shouldn't use it here either. just use the entry.term as the key

src/app/utils/abbreviations.ts
```diff
@@ -0,0 +1,61 @@
...
+  const terms = [...abbrMap.keys()].sort((a, b) => b.length - a.length);
+  const escaped = terms.map((t) => t.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'));
+  const pattern = new RegExp(`\\b(${escaped.join('|')})\\b`, 'gi');
+
+  const segments: TextSegment[] = [];
+  let lastIndex = 0;
+  let match: RegExpExecArray | null;
+
+  // eslint-disable-next-line no-cond-assign
+  while ((match = pattern.exec(text)) !== null) {
```
7w1: Could replace this loop with this so we dont need to disable eslint?
```typescript
for (const match of text.matchAll(pattern)) {
  const matchIndex = match.index!;
```

7w1 (changes requested):

7w1 (approved):

PR #516, opened 2026-03-24
fix: prevent 401 errors on authenticated media after service worker restart

Just-Insane:
### Description

The service worker (SW) loses its in-memory session on restart (e.g. after the browser kills the SW background process). Any authenticated media request arriving before the SW re-reads session data from cache was returning a 401.

Two fixes:
1. **Eager pre-load on `activate`**: the SW now reads session data from the cache during the `activate` event, so it is available immediately when the first `fetch` event fires.
2. **Graceful fallback in fetch handler**: if session data is still missing or stale at request time, the request is passed through to the network without injecting a bad `Authorization` header (previously an empty/invalid token was injected, causing the 401).

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

Attempts to resolve issues with the service worker not properly storing and/or refreshing media auth tokens on load. The changes mainly serve to ensure that a session is persisted during restarts (and then drop the persisted session once a live one is available). Also helps with refreshing tokens (`sw.ts` ln 651-661)

### DIFF
--- a/.changeset/fix-sw-media-auth-401.md
+++ b/.changeset/fix-sw-media-auth-401.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix service worker authenticated media requests returning 401 errors after SW restart or when session data is missing/stale.
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -460,6 +460,11 @@ self.addEventListener('activate', (event: ExtendableEvent) => {
     (async () => {
       await self.clients.claim();
       await cleanupDeadClients();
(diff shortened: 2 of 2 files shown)


Just-Insane:
~~This may actually cause issues with resuming from background in iOS - it seems like new messages don't show in the timeline until the PWA is closed. More testing required.~~

Edit:

Actually, I think the messages are just rendering out of order (which I don't think is related to this PR) - Clicking on a message notification brings me to the message, at the top of the synced message group.

Should be fixed by 1. in https://github.com/SableClient/Sable/pull/529

7w1 (approved):

PR #524, opened 2026-03-24
Feat: add quickreply keybindings

CodeF53:
### Description

This adds the keybinds `ctrl+up`/`ctrl+down` to be used to cycle what message you are replying to. This is inspired by [the vencord plugin QuickReply](https://vencord.dev/plugins/QuickReply)

#### Type of change

### Checklist:

- [x]  I have performed a self-review of my own code

### AI disclosure:

I couldn't get `pnpm dev ` to run so I used an llm to help me figure that out so I could test my changes, but no code in this pr has been touched by an llm

### DIFF
--- a/src/app/components/GlobalKeyboardShortcuts.tsx
+++ b/src/app/components/GlobalKeyboardShortcuts.tsx
@@ -5,10 +5,11 @@
  *   Alt+N              — jump to the highest-priority unread room
  *   Alt+Shift+Down     — cycle forward through unread rooms
  *   Alt+Shift+Up       — cycle backward through unread rooms
+ *   Ctrl+Down / Ctrl+Up: cycle through messages to reply to
  */
 import { useCallback, useRef } from 'react';
 import { useNavigate, useLocation, matchPath } from 'react-router-dom';
-import { useAtomValue } from 'jotai';
+import { useAtomValue, useSetAtom } from 'jotai';
 import { isKeyHotkey } from 'is-hotkey';
 import { useMatrixClient } from '$hooks/useMatrixClient';
 import { roomToParentsAtom } from '$state/room/roomToParents';
@@ -19,6 +20,8 @@ import { getDirectRoomPath, getHomeRoomPath, getSpaceRoomPath } from '$pages/pat
 import { HOME_ROOM_PATH, DIRECT_ROOM_PATH, SPACE_ROOM_PATH } from '$pages/paths';
 import { getCanonicalAliasOrRoomId } from '$utils/matrix';
 import { announce } from '$utils/announce';
+import { roomIdToReplyDraftAtomFamily } from '$state/room/roomInputDrafts';
+import { Room } from 'matrix-js-sdk';
 
 export function GlobalKeyboardShortcuts() {
   const navigate = useNavigate();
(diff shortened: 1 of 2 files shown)


src/app/components/GlobalKeyboardShortcuts.tsx
```diff
@@ -37,15 +39,17 @@ export function GlobalKeyboardShortcuts() {
   const roomIdOrAlias = roomMatch?.params.roomIdOrAlias
     ? decodeURIComponent(roomMatch.params.roomIdOrAlias)
     : undefined;
-  let currentRoomId: string | null = null;
-  if (roomIdOrAlias) {
+  let currentRoom: Room | null = null;
+  if (roomIdOrAlias !== undefined) {
     if (roomIdOrAlias.startsWith('!')) {
-      currentRoomId = roomIdOrAlias;
+      currentRoom = mx.getRooms().find((r) => r.roomId === roomIdOrAlias) ?? null;
```
CodeF53: I dont understand enough about sable's room ids and routing params to understand what this code path is even supposed to handle, so I wasnt able to test it

CodeF53:
https://github.com/user-attachments/assets/bbe3dd23-1e7b-40fd-9140-bde52e4d7760

I feel like sable missing the highlight for the message you are replying to makes it harder to use this than it is on discord, but adding a highlight for the reply target is out of scope for this pr in my opinion

dozro:

Thank you :3

src/app/components/GlobalKeyboardShortcuts.tsx
dozro, line 7: to keep the file heading a bit up to date :3

```suggestion
 *   Alt+Shift+Up       — cycle backward through unread rooms
 *   Ctrl+Down / Ctrl+Up: cycle through messages to reply to
```

src/app/components/GlobalKeyboardShortcuts.tsx
dozro:
```suggestion
      currentRoom = mx.getRoom(roomIdOrAlias) ?? null;
```

dozro (changes requested):

src/app/components/GlobalKeyboardShortcuts.tsx
```diff
@@ -37,15 +39,17 @@ export function GlobalKeyboardShortcuts() {
   const roomIdOrAlias = roomMatch?.params.roomIdOrAlias
     ? decodeURIComponent(roomMatch.params.roomIdOrAlias)
     : undefined;
-  let currentRoomId: string | null = null;
-  if (roomIdOrAlias) {
+  let currentRoom: Room | null = null;
+  if (roomIdOrAlias !== undefined) {
```
dozro: i think that change is unnecessary and changes behaviour (potentially). The `if(roomIdOrAlias!== undefined)` is less restrictive than `if(roomIdOrAlias)`.

`roomIdOrAlias !== undefined` is true when `roomIdOrAlias` is an empty string for example, unsure if we want that here.


```suggestion
  if (roomIdOrAlias) {
```

CodeF53: ~~`!== undefined` is significantly more explicit so should be preferred and if you have an issue with it you should refactor this pattern out of the codebase the 177 times it occurs already~~ this was far more aggressive sounding than I intended it to be
image
image
image
image

CodeF53: > roomIdOrAlias !== undefined is true when roomIdOrAlias is an empty string for example, unsure if we want that here.

wait what why on earth do the strict operators make `"" === undefined` that is just blatantly bad behavior for what should be a stricter check

CodeF53: wait no I am just not thinking at all

CodeF53: is there a way we can have a stricter check while maintaining existing behavior, I am really not a fan of `if (var)` when `var` is not a boolean

CodeF53:
Sorry about all the force pushes, nixos is really strict about one commit per file changed in a pr and it has formed some habits in me

src/app/components/GlobalKeyboardShortcuts.tsx
CodeF53: I dont like `roomIdOrAlias !== undefined || roomIdOrAlias === ''` so I did just what was suggested

7w1: Could just make it a boolean :3
if (!!var)
/hj

dozro: > is there a way we can have a stricter check while maintaining existing behavior, I am really not a fan of `if (var)` when `var` is not a boolean

maybe `if(roomIdOrAlias?.length > 0)` or smth

7w1: Otherwise… typeof === “string” and a length check?

dozro (approved):
code lgtm

7w1 (approved):

PR #535, opened 2026-03-25
fix confusing ui for client side embed settings

thundertheidiot:
### Description

this is a tiny fix to a potentially confusing "problem" that i thought of when going to sleep last night, previously if you enabled client side embeds in both unencrypted and encrypted rooms, and then turned off the main client side embeds toggle, the encrypted rooms setting would remain on, even though it would be hidden in the setting ui

feel free to just close if this is stupid

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/.changeset/fix_confusing_ui_with_client_side_embeds_setting.md
+++ b/.changeset/fix_confusing_ui_with_client_side_embeds_setting.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Fix confusing ui with `Client Side Embeds in Encrypted Rooms` setting
--- a/src/app/features/room/RoomTimeline.tsx
+++ b/src/app/features/room/RoomTimeline.tsx
@@ -145,7 +145,7 @@ export function RoomTimeline({
 
   const showUrlPreview = room.hasEncryptionStateEvent() ? encUrlPreview : urlPreview;
   const showClientUrlPreview = room.hasEncryptionStateEvent()
-    ? encClientUrlPreview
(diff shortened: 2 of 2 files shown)


dozro:
does the UI in the settings currently allow to set the setting for embeds in encrypted rooms to `on` while embeds for normal rooms are `off`?

if so, we should maybe change that… something like the thing with hidden events and tombstones, or maybe only show the option when you also selected the other 🤔

thundertheidiot:
the client embeds (unencrypted rooms) toggle toggles the visibility of the other settings under it, but you could turn both on, and then disable the unencrypted one, which would hide the other settings, but actually secretly leave the encrypted room embed setting on, which felt pretty unintuitive

this fixes that, so that if you turn off the basic toggle, also the encrypted rooms support is effectively disabled

dozro:
hmmm that should probably be fixed at the source as well then. iirc correctly from first looking at this Pr lol, it changes it downstream 🤔

thundertheidiot:
i mean i think it's fine for the encrypted toggle to remain on, if you want to quickly just disable all client embeds for a bit and then turn them back on later, it would make sense for the encrypted ones to stay on so you don't have to click on two things later?
i might be stupid

dozro:
i think that’s not the behavior a user expects, but i don’t know tbh

7w1 (approved):

7w1:
Might be worth just graying out the encrypted toggle if the normal toggle is disabled like we do for a few other settings I can't remember. I know we do that with the desktop only new line setting.

PR #538, opened 2026-03-25
fix(unread): fix badge counts stuck at "1" for unvisited rooms in sliding sync

Just-Insane:
### Description

With sliding sync, unvisited rooms (no read receipt) show a badge of **1** instead of the true unread count. After visiting a room, the badge corrects itself. After a cache clear or force-close, all badges are wrong on the initial sync pass.

**Root Cause**

Room list subscriptions use `timeline_limit=1`, so unvisited rooms only have a single event in memory. `room.fixupNotifications(userId)` re-derives the notification count from whatever events are in the live timeline — for these rooms that's exactly 1 event, so it sets `_notificationCounts.total = 1` and overwrites the server-supplied `notification_count`.

Flow for a room with 15 true unreads:
1. Server delivers `notification_count: 15` → stored in SDK
2. `fixupNotifications()` → recalculates from 1-event timeline → `total = 1`
3. `roomHaveUnread()` returns `false` (no read receipt) → stale-clamp fires but `hasUserReadEvent` is also false → total stays at 1
4. `!readUpToId` fallback sees `hasActivity=true, total=1≠0` → returns `{ total: 1 }` ← bug

After the user opens a room, the active subscription delivers the full timeline and a read receipt is sent. At that point `fixup` gets the full event set and works correctly.

**Fix**

Guard `room.fixupNotifications(userId)` with `room.getEventReadUpTo(userId)` so fixup is only applied for rooms the user has visited (where a read receipt or fully-read marker is present). Unvisited rooms use the raw server-supplied `notification_count` directly.

This also resolves the related symptom where unread counts are wrong after a cache clear or force-close: all rooms start without receipts, so fixup was corrupting every badge on the initial sync pass.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

Adds `room.getEventReadUpTo(userId)` guard around the `room.fixupNotifications(userId)` call in `getUnreadInfo` in `src/app/utils/room.ts`. When the guard fails (no read receipt present), the function skips the fixup and uses the server-provided notification count directly.

### DIFF
--- a/.changeset/fix-unread-counts.md
+++ b/.changeset/fix-unread-counts.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix unread badge counts showing "1" for unvisited rooms by guarding fixupNotifications to only run for rooms where the user has a read receipt.
--- a/src/app/utils/room.ts
+++ b/src/app/utils/room.ts
@@ -291,13 +291,18 @@ type UnreadInfoOptions = {
 export const getUnreadInfo = (room: Room, options?: UnreadInfoOptions): UnreadInfo => {
   const userId = room.client.getUserId();
   if (userId && options?.applyFixup) {
-    // Only apply fixup for rooms the user has visited (read receipt / fully-read marker
-    // present). For unvisited rooms, fixupNotifications() re-derives the notification count
-    // from the limited sliding sync list timeline (timeline_limit=1, so 1 event) and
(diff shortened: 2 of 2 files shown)


dozro:

Thank you :3

Just-Insane:
Updated this slightly so that it works better for sliding sync

Just-Insane:
I can't get this reliably working with the current sliding sync implementation. Will leave this open (draft) and come back to it at some point maybe.

PR #548, opened 2026-03-26
fix: intermittent media 401 errors after token refresh or SW restart

Just-Insane:
### Description

Fixes a chain of service worker session bugs that caused intermittent 401/404 errors on authenticated media:

- Media requests no longer return 401 after a token refresh or SW restart
- Session cache is now guaranteed to flush before the SW can be killed, fixing intermittent 401s on iOS
- OIDC token refreshes are forwarded to the service worker as they happen
- Fixed unhandled promise rejections from media decrypt failures

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

- [x] Partially AI assisted (GitHub Copilot)

`sw.ts`: The `setSession`/`clearSession` cache writes are now wrapped in `event.waitUntil` so the browser keeps the SW alive until the `caches.put` resolves — preventing stale tokens surviving a mid-write SW kill on iOS. Added a `requestSessionWithTimeout` fallback in `fetchMediaWithRetry` for the window between a token refresh completing and the resulting `setSession` message arriving. `index.tsx`: Calls `pushSessionToSW` synchronously before React renders so the SW has a valid token before the first fetch event. `ClientRoot.tsx`: Wires `pushSessionToSW` into the SDK's `onTokenRefresh` callback so OIDC token refreshes propagate to the SW immediately. Decrypt failure and media error async paths now have proper rejection handling to avoid unhandled promise rejections.

### DIFF
--- a/.changeset/fix-media-error-handling.md
+++ b/.changeset/fix-media-error-handling.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix intermittent 401 errors when loading media after service worker restart. Service worker now expires cached authentication tokens after 60 seconds, forcing fresh token retrieval from the active page instead of using potentially stale persisted tokens.
--- a/Caddyfile
+++ b/Caddyfile
@@ -16,6 +16,13 @@
 
 	try_files {path} /index.html
 
+	# Content-hashed assets can be cached indefinitely — the filename changes on every build.
+	header /assets/* Cache-Control "public, max-age=31536000, immutable"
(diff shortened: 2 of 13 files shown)


7w1:
I think there might be some extraneous features in here? The whole new dev tools section, the UrlPreviewCard.tsx normalization changes and the sliding sync probe caching? 

I think this should either be split into multiple prs or the description should be updated and there should be more changesets. In general I think the code looks fine.

7w1 (changes requested):
update changesets or split into multiple prs

PR #552, opened 2026-03-26
Feat: Add preventing urlPreviewing by surrounding a link in angle brackets

nushea:
### Description
This is an implementation with the main purpose to serve as a proposal.

Currently there is no way to tell the client to not display a urlPreview if they are enabled, this PR proposes adding it by surrounding the link in angle brackets such as `<https://app.sable.moe>`, `[Sable!](<https://app.sable.moe>)`, or `<[Sable!](https://app.sable.moe)>`
The message saves the angle brackets outside of the link tag regardless to not break the links on any other client that does not have the regular hide preview mode (so regardless of whether you type `[foo](<https://bar>)` or `<[foo](https://bar)>` it will save it as `<<a data-md href='https://bar'>foo</a>>` as formatted text and `<(foo)[https://bar]>` as regular text)

I am not sure if this would count as proper markdown but it is a very useful feature on discord and order platforms that have urlPreviews in order to prevent irrelevant previews intruding in the timeline.

Additionally, the implementation is working for me in all the cases i have tested but i am not a regex person by any means so if anyone could doublecheck my work it would be greatly appreciated.

Lastly, in this implementation the angle brackets get hidden regardless of whether it is rendering the formatted_body or the body, but i am not 100% sure if the regular body should get this type of formatting.

(normal look)
image

(new look)
image

#### Type of change

### Checklist:

### AI disclosure:

A demon whispered the implementation in my ears.

### DIFF
--- a/.changeset/add_hide_urlPreview.md
+++ b/.changeset/add_hide_urlPreview.md
@@ -1,0 +1,5 @@
+---
+default: minor
+---
+
+Add preventing urlPreviewing by surrounding a link in angle brackets
--- a/src/app/components/message/MsgTypeRenderers.tsx
+++ b/src/app/components/message/MsgTypeRenderers.tsx
@@ -139,16 +139,22 @@ export function MText({ edited, content, renderBody, renderUrlsPreview, style }:
 
   if (!body && !customBody) return <BrokenContent body={customBody ?? body} />;
 
+  // move the < > inside of the []() in order to make the matching easier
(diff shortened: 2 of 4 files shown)


hazre:
Good stuff, my only feedback would be to lower the scope a bit and only allow `<>` around urls, not markdown itself. That should simply the code a lot more and should be more inline with something like discord.

So in case you want to use it in a link markdown, it can be used like this (as you already suggested):

`[Sable](<https://app.sable.moe>)`

Instead of this

`<[Sable](https://app.sable.moe)>`

PR #572, opened 2026-03-28
feat(flags): build-time client config overrides via Feature Flags

Just-Insane:
📝 **Docs PR:** SableClient/docs#8 — merge that when merging this.

---

### Description

Adds build-time runtime-config override support so config.json can be controlled per GitHub Environment without source edits. This is the shared infrastructure for staged rollout experiments.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

The injector script and workflow wiring were partially AI assisted. I reviewed deep-merge behavior, failure modes, and no-op safety when vars are absent.

### What this PR adds

- Build-time config injector script that merges environment overrides into config.json
- Workflow wiring so preview and production use their own GitHub Environment variables
- Typed experiment selection helpers for deterministic bucketing
- Safe fallback behavior when environment variables are missing

### Why this implementation is useful

- Removes code-change churn for rollout changes. Product/ops can adjust percentages and experiment toggles by environment variables instead of opening code PRs for every percentage change.
- Enables safer progressive delivery. Preview can run at 100% while production starts low and ramps gradually.
- Supports consistent experiment assignment. Deterministic bucketing keeps users stable in control/treatment cohorts across sessions.
- Reduces deployment risk. If override variables are missing, malformed, or intentionally disabled, builds still fall back to checked-in defaults.
- Improves incident response. Feature rollbacks become a config operation rather than a code revert/redeploy cycle.
- Keeps app code and release controls separated. The app reads typed config while CI/environment owns operational rollout state.

### Environment variables and what they do

Required variable:

- `CLIENT_CONFIG_OVERRIDES_JSON` — JSON object deep-merged into config.json at build time
- `CLIENT_CONFIG_OVERRIDES_STRICT` — set to `"true"` to fail the build on malformed JSON rather than warn

### DIFF
--- a/.changeset/feature-flag-env-vars.md
+++ b/.changeset/feature-flag-env-vars.md
@@ -1,0 +1,5 @@
+---
+default: minor
+---
+
+Add build-time client config overrides via environment variables, with typed deterministic experiment bucketing helpers for progressive feature rollout and A/B testing.
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -34,6 +34,36 @@ runs:
       env:
         INPUTS_INSTALL_COMMAND: ${{ inputs.install-command }}
 
+    - name: Inject runtime config overrides
+      if: ${{ env.CLIENT_CONFIG_OVERRIDES_JSON != '' }}
+      shell: bash
+      working-directory: ${{ github.workspace }}
+      run: node scripts/inject-client-config.js
+      env:
+        CLIENT_CONFIG_OVERRIDES_JSON: ${{ env.CLIENT_CONFIG_OVERRIDES_JSON }}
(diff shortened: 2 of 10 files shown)


Just-Insane:
Implemented the Copilot review feedback in commit `6c4e6672`:

1. Added dedicated unit tests for experiment selection edge cases in `src/app/hooks/useClientConfig.test.ts`.
2. Fixed injector logging in `scripts/inject-client-config.js`:
   - removed call to missing `logger.warn(...)`
   - ensured errors include exception details via formatted error strings

Validated with:
- `pnpm vitest run src/app/hooks/useClientConfig.test.ts` (8 passing tests)

If you want, I can also split the strict/non-strict malformed-JSON paths into script-level tests in a follow-up.

PR #659, opened 2026-04-10
perf(timeline): restore VList cache + skip 80 ms timer on room revisit

Just-Insane:
### Description

Caches VList item heights (via `roomScrollCache`) across visits to the same room within a session. On revisit, the stored cache is provided to VList upfront and the scroll position is restored immediately — skipping the 80 ms opacity-fade stabilisation timer entirely.

Also fixes a bug where the `pendingReadyRef` recovery path (triggered when the 80 ms timer fires while `processedEvents` is empty due to a `TimelineReset` race) never saved the scroll cache, causing rooms that hit this path to show the fade-in on subsequent visits.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

`roomScrollCache.ts` — a simple session-scoped `Map<roomId, { cache, scrollOffset, atBottom }>`. `cache` is VList's `CacheSnapshot` (measured item heights), `scrollOffset` is the pixel offset at save time, and `atBottom` is a flag used to decide whether to call `scrollToIndex` (bottom) or `scrollTo` (exact offset) on restore.

In `RoomTimeline.tsx`, the initial-scroll `useLayoutEffect` was extended to check for a saved cache before starting the 80 ms timer: if a cache exists, it restores position immediately and calls `setIsReady(true)` without the delay. The 80 ms timer path and the `pendingReadyRef` recovery path both now call `roomScrollCache.save(...)` after scrolling so subsequent visits have a cache to load.

### DIFF
--- a/.changeset/perf-timeline-scroll-cache.md
+++ b/.changeset/perf-timeline-scroll-cache.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Cache VList item heights across room visits to restore scroll position instantly and skip the 80 ms opacity-fade stabilisation timer on revisit.
--- a/src/app/features/room/RoomTimeline.tsx
+++ b/src/app/features/room/RoomTimeline.tsx
@@ -316,6 +316,9 @@ export function RoomTimeline({
         // immediately and skip the 80 ms stabilisation timer entirely.
         if (savedCache.atBottom) {
           vListRef.current.scrollToIndex(processedEventsRef.current.length - 1, { align: 'end' });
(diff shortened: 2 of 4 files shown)


Just-Insane:
Added a follow-up fix in the latest commit: `setAtBottom(true)` is now called synchronously when scrolling to the bottom in the cache-restore and `pendingReadyRef` recovery paths, immediately before `setIsReady(true)`. This prevents the "Jump to Latest" button from flashing for one render cycle between `isReady` becoming `true` and VList's async `onScroll` confirming the bottom position.

PR #660, opened 2026-04-10
perf(timeline): memoize VList timeline items to prevent mass re-renders

Just-Insane:
### Description

Wraps each VList timeline item in `React.memo` so that only the items whose props actually changed re-render when state updates occur (e.g. typing indicators arriving, read receipts updating, or new messages landing while the user is scrolled away from the bottom).

Props passed to each item beyond `data` and `renderRef` — `isHighlighted`, `isEditing`, `isReplying`, `isOpenThread`, `settingsEpoch` — exist solely to give `React.memo`'s shallow-equality comparator fine-grained control: changing any of them causes only the one affected item to re-render rather than the entire visible list.

Also makes `mutationVersion` optional (default `0`) in `UseProcessedTimelineOptions` so call sites that don't participate in the mutation tracking (e.g. `ThreadDrawer`) continue to compile without changes.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

`useProcessedTimeline.ts` — added a `stableRefsCache` (a `useRef<Map<string, ProcessedEvent>>`) that keeps the previous render's `ProcessedEvent` objects alive. On each `useMemo` run, if `mutationVersion` hasn't changed, events whose identity hasn't changed are returned from the cache (same object reference) rather than newly constructed. This means `React.memo` sees no prop change for stable items and skips their render.

`RoomTimeline.tsx` — `TimelineItem` is a `React.memo`-wrapped component declared outside the parent function so it is never re-created. A `renderFnRef` ref captures the render function so `TimelineItem` itself only needs a stable ref in props, not the function. `settingsEpoch` is a `useRef({}).current` that is replaced with `{}` whenever any layout/display setting changes, forcing all visible items to pick up the new settings in one pass.

### DIFF
--- a/.changeset/perf-timeline-item-memo.md
+++ b/.changeset/perf-timeline-item-memo.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Memoize individual VList timeline items to prevent mass re-renders when unrelated state changes (e.g. typing indicators, read receipts, or new messages while not at the bottom).
--- a/.changeset/perf-timeline-scroll-cache.md
+++ b/.changeset/perf-timeline-scroll-cache.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Cache VList item heights across room visits to restore scroll position instantly and skip the 80 ms opacity-fade stabilisation timer on revisit.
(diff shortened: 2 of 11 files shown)


Just-Insane:
Added a follow-up fix in the latest commit: `setAtBottom(true)` is now called synchronously when scrolling to the bottom in the 80ms timer and `pendingReadyRef` recovery paths, immediately before `setIsReady(true)`. This prevents the "Jump to Latest" button from flashing for one render cycle between `isReady` becoming `true` and VList's async `onScroll` confirming the bottom position.

PR #666, opened 2026-04-12
feat(dm-list): show last-message preview below room name

Just-Insane:
### Description

Adds a real-time last-message preview (most recent non-state, non-reaction event) to each DM room entry in the sidebar, mirroring what most messaging clients show in their room list.

**Commits in this PR:**

1. **`feat(dm-list): add toggle to hide DM message preview`** — adds the feature and a Visual Tweaks toggle to disable it. The preview renders the latest timeline event using a format suited to each content type (text, attachment, sticker, poll, etc.) and avoids re-rendering unless the relevant timeline events or atoms actually change.

2. **`feat(dm-list): prefix message preview with sender display name`** — prepends the sender's display name to the preview string (e.g. "Alice: Hello!") so the conversation thread is clearer at a glance without opening the room.

3. **`fix(settings): give DM Message Preview its own card in Visual Tweaks`** — moves the toggle into its own dedicated settings card so it doesn't get visually buried with unrelated options.

4. **`fix(dm-list): update message preview immediately on event decryption`** — subscribes to `MatrixEventEvent.Decrypted` so the preview Updates the moment an encrypted event is decrypted, preventing a stale "Encrypted message" placeholder from lingering after decryption.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

The `useRoomLastMessage` hook's reverse-scan loop, the content-type dispatch table, and the `MatrixEventEvent.Decrypted` listener pattern were drafted with AI assistance and reviewed against the Matrix JS SDK documentation and existing hook patterns in the codebase. The settings card restructuring and sender-prefix logic were written manually.

### DIFF
--- a/.changeset/feat-dm-message-preview.md
+++ b/.changeset/feat-dm-message-preview.md
@@ -1,0 +1,5 @@
+---
+'@sable/client': minor
+---
+
+feat(dm-list): show last-message preview below DM room name
--- a/src/app/features/room-nav/RoomNavItem.tsx
+++ b/src/app/features/room-nav/RoomNavItem.tsx
@@ -265,6 +265,7 @@ type RoomNavItemProps = {
   showAvatar?: boolean;
   direct?: boolean;
   customDMCards?: boolean;
+  dmMessagePreview?: boolean;
 };
 
 export function RoomNavItem({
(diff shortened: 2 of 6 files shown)


nushea:
i think maybe this should be a setting (and maybe, if the message is undecrypted it should fall-back to the status)

Just-Insane:
> i think maybe this should be a setting (and maybe, if the message is undecrypted it should fall-back to the status)

It is a toggle-able setting, and if the setting to show the room description is enabled, it overrides the message preview.

PR #677, opened 2026-04-13
Add GIF search feature

### LINKED ISSUE #60: GIF Search
### Describe the problem

https://github.com/cinnyapp/cinny/pull/2392

This, but it should use https://klipy.com/ since Tenor has been discontinued. Several other clients and major open source projects have moved to Klipy.

Enable GIF Search with Klipy as a provider.

### Alternatives considered

_No response_

### Additional context

_No response_

footvaalvica:
### Description

Fixes #60 after all TODOs below have been completed. 

#### Open Problems
- The backend currently works with Commet's Tenor proxy. It should be migrated to KLIPY as per the original issue.
- The icon currently being used is Icons.Play. It should be replaced with a proper GIF icon.

The approach for sending GIFs is to download the GIF data and then re upload them as a sticker.

This PR was heavily inspired by https://github.com/cinnyapp/cinny/pull/2392, so if possible @razn-v should give their input and be added as a co-contributor.

#### Type of change

### Checklist:

### AI disclosure:

The changes to renderItem() to support GIFs were AI generated. It returns GIF Items and ensures they are scaled properly.


dozro:

Thank you :3

footvaalvica:
thanks for the heads up, I completely missed it. should be done now!

src/app/features/room/RoomInput.tsx
```diff
@@ -1140,6 +1140,57 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
...
+      }
+
+      const data = await response.arrayBuffer();
+      const uint8Array = new Uint8Array(data);
+
+      // Create a File object for the GIF
+      const filename = `${gif.title}.gif`;
+      const file = new File([uint8Array], filename, { type: 'image/gif' });
+
+      // Upload to Matrix
```
dozro, line 1175: up for debate imo is, if we want to reupload the gif to the hs, or use something like https://github.com/maunium/stickerpicker/tree/master/giphyproxy 

(as discussed, but afaik not resolved in the sable client matrix room)

.gitignore
```diff
@@ -23,3 +23,24 @@ build.sh
...
+.envrc
+devenv.lock
+
+# direnv
+.direnv
+
+# pre-commit
+.pre-commit-config.yaml
+CLAUDE.md
+.github/AGENTS.md
```
dozro: ?

hazre:
I think if we going to add this, might as well go with KLIPY, right off the bat. You can get a dev key on their website and later we can contact for a prod key. It seems to be free and we just have to respect their API rate limits.

hazre:
Unless you doing server level caching, I don't see a reason for a proxy in this case. I guess you can avoid shipping the secrets in the client.

[Commet's proxy](https://github.com/commetchat/proxy) is just a cloudflare worker, which is fine but we are currently on the free tier of cloudflare which has 100k daily requests limit (if we were to host it ourselves)

@7w1 thoughts?

dozro:
> Unless you doing server level caching, I don't see a reason for a proxy in this case. I guess you can avoid shipping the secrets in the client.

The homeserver by default does caching for remote media. Which would be utilized with that proxy. And cached remote media is easily purgeable, while re-uploaded media is harder to manage and will easily get out of hand. Especially for bigger Homeservers the way to go would be the proxy, it would by its nature alone solve the issue of duplicated gifs taking up too much space in the media repository.


btw comments proxy does afaict way differs things than the maunium sticker picker proxy

7w1:
Maunium's proxy makes sense to me since it essentially acts like a homeserver and allows other homeservers to cache data and embed smoothly without needing to be reuploaded by everyone.

dozro:
> Maunium's proxy makes sense to me since it essentially acts like a homeserver and allows other homeservers to cache data and embed smoothly without needing to be reuploaded by everyone.

it could use https://github.com/SableClient/soliditas it isn’t tested that throughly tho, so bugs can exist 😅

7w1:
Superseded by #970

PR #696, opened 2026-04-17
Migrate to oxlint and oxfmt

7w1:
### Description

Migrates from eslint and prettier to oxlint and oxfmt + some rule hardening and - some rules that don't exist in oxlint, but I decided it probably wasn't worth running both in parallel just for a few rules.

Notably, got rid of many "as any"s and swapped to .toSorted() for safer stuff and "type" in imports. Added comments to all the existing lint issues (console logging and the alert for app updates) and the few places "as any" was justified (test cases) and some message object whatnots.

Also notably, fmt now takes a second or two, and linting now takes no more than a few seconds. Yay!

Recommended vscode plugins, workflows, and configs should all be updated assuming I didn't miss anything.

#### Type of change

- [x] Breaking change - breaking I guess since not all rules exist? I don't really know exactly what I uh... didn't include.

### Checklist:

### AI disclosure:

I deployed a swarm of silly AIs to fix many of the simpler as any casts and subsequent typechecks because there were hundreds and I'm not doing that manually :p. I did review it, mostly just seamless swapping to the correct cast type instead of as any, nothing particularly complicated.


nushea:
read it partially buton the train on my phone  the code lgtm for the parts I was concerned about

hazre (commented):
Oh nice, I was actually going to do this in #607. 

Few things, I think we should use the *.config.ts format instead of json, since it's more flexible.
https://oxc.rs/docs/guide/usage/linter/config.html#typescript-config-file-oxlint-config-ts
https://oxc.rs/docs/guide/usage/formatter/config.html#create-a-config-file


Also move fix.py to scripts and ideally rewrite in js and rename it to be less generic.

.oxlintrc.json
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "./node_modules/oxlint/configuration_schema.json",
+  "plugins": ["react", "jsx-a11y", "typescript", "import"],
```
hazre: Also add type aware linting
https://oxc.rs/docs/guide/usage/linter/type-aware.html

also probably should add `vitest`, `react-perf`, `promise` plugins

I think `unicorn`, `oxc` is included by default but probably should be explicit about it.

hazre (changes requested):

hazre (changes requested):
Looks good. Any chance you could also setup these:
https://oxc.rs/docs/guide/usage/linter/editors.html#team-setup
https://oxc.rs/docs/guide/usage/formatter/editors.html#team-setup

for automatic linting and formatting on save?

also could you remove stale eslint comments or replace them with oxlint ones?

for consistency sakes swapping oxfmt to config.ts would be nice too.

hazre:
@7w1 I think this is good to go, wanna take a look at my changes and merge it?

PR #743, opened 2026-05-05
Fix of wrong alignment with mixed RTL and LTR text

sinasadeghi83:
### Description

This PR resolves formatting and alignment issues with mixed RTL (Right-to-Left) and LTR (Left-to-Right) text within messages. Previously, mixed text direction caused display inconsistencies in the main message body and the reply content area.

**Changes made:**
* Added `dir="auto"` to the base message `Text` component, allowing the browser to automatically determine the base text direction according to the first strongly-typed character of the message.
* Applied `unicode-bidi: plaintext` to the message content CSS and the `Text` component inside the reply content. This ensures the browser dynamically evaluates and handles the text direction based on the actual characters used rather than inheriting the container's direction.
* Set `align-self: start` on the message content. This ensures that even for RTL messages, the content block anchors appropriately to the left side (start) of the view, while still preserving the correct internal text direction.

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/.changeset/fix_rtlltr_mixed_text_formatting_and_alignment_in_messages.md
+++ b/.changeset/fix_rtlltr_mixed_text_formatting_and_alignment_in_messages.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Fix RTL/LTR mixed text formatting and alignment in messages
--- a/src/app/components/message/Reply.tsx
+++ b/src/app/components/message/Reply.tsx
@@ -280,7 +280,7 @@ export const Reply = as<'div', ReplyProps>(
           onClick={replyEvent !== null && !isBlockedSender ? onClick : undefined}
         >
           {replyEvent !== undefined && !isPendingDecrypt ? (
-            <Text size="T300" truncate>
+            <Text size="T300" truncate style={{ unicodeBidi: 'plaintext' }}>
               {replyContent}
             </Text>
           ) : (
--- a/src/app/components/message/layout/Base.tsx
+++ b/src/app/components/message/layout/Base.tsx
@@ -48,6 +48,7 @@ export const MessageTextBody = as<'div', css.MessageTextBodyVariants & { notice?
       className={classNames(css.MessageTextBody({ preWrap, jumboEmoji, emote }), className)}
       {...props}
       ref={ref}
+      dir="auto"
     >
       {children}
     </Text>
--- a/src/app/components/message/layout/layout.css.ts
+++ b/src/app/components/message/layout/layout.css.ts
@@ -229,6 +229,8 @@ export const PronounPill = style({
 
 export const MessageTextBody = recipe({
   base: {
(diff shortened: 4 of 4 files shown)


7w1:
Hey, thanks for contributing! Could you write a changeset as described [here](https://github.com/SableClient/Sable/blob/dev/CONTRIBUTING.md#documenting-a-change)?

AmirAref:
Just as a user who needed this feature a lot, first I want to say thank you.
I have a suggestion to consider, wouldn't it be better to display RTL text starting from the right side of the row? I think it would be easier to read for RTL language readers (like Persian).
Here is an example of what I mean:

Current state:
swappy-20260508-143446



RTL text aligned to the right:
swappy-20260508-143709

@sinasadeghi83

hazre:
I like @AmirAref's suggestion

sinasadeghi83:
> Just as a user who needed this feature a lot, first I want to say thank you. I have a suggestion to consider, wouldn't it be better to display RTL text starting from the right side of the row? I think it would be easier to read for RTL language readers (like Persian). Here is an example of what I mean:
> 

It's against proximity principle in visual design. Texts can't be too far from profile name and profile image because then you have to roll your eyes all to the left to see who sent the message. Think about this scenario when you are reading a chain of messages from Alice then suddenly Bob sends a new message, then at first you may assume that Alice is the sender because you haven't seen the sender name or profile image yet. Then after reading the whole message to the end, you find out that Alice couldn't be the sender. You roll your eyes to the left and see that oh it's Bob!
But when message body is close to profile name and image, you would see sender before their message content.
Other messaging apps follow the same design too.

sinasadeghi83:
> Hey, thanks for contributing! Could you write a changeset as described [here](https://github.com/SableClient/Sable/blob/dev/CONTRIBUTING.md#documenting-a-change)?

You're welcome. It's done.

AmirAref:
> > Just as a user who needed this feature a lot, first I want to say thank you. I have a suggestion to consider, wouldn't it be better to display RTL text starting from the right side of the row? I think it would be easier to read for RTL language readers (like Persian). Here is an example of what I mean:
> 
> It's against proximity principle in visual design. Texts can't be too far from profile name and profile image because then you have to roll your eyes all to the left to see who sent the message. Think about this scenario when you are reading a chain of messages from Alice then suddenly Bob sends a new message, then at first you may assume that Alice is the sender because you haven't seen the sender name or profile image yet. Then after reading the whole message to the end, you find out that Alice couldn't be the sender. You roll your eyes to the left and see that oh it's Bob! But when message body is close to profile name and image, you would see sender before their message content. Other messaging apps follow the same design too.

Yeah, you're right. I think this would be handled greatly by changing message layout to **Bubble** and enable **_Right Aligned Bubbles_**  in General settings of Sable.

7w1 (approved):

PR #798, opened 2026-05-13
Improve caption behavior

7w1:
### Description

Still need to figure out what behaviors are the best for this.

Adjusted captions so the alt text on an image is set when captions are hidden, and is otherwise blank, to improve screen reader behavior.

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/.changeset/fix-alt-text-behavior.md
+++ b/.changeset/fix-alt-text-behavior.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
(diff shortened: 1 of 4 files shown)


RoootTheFox:
worth mentioning that orca seems to always read image captions; but only read text if its selected through caret browsing (shift+arrow keys) - in other words, always setting alt text might be a good idea.

waiting for opinions on this from people who actually rely on screen readers would also be a good idea ^^

PR #861, opened 2026-05-18
Feat add basic hiding rooms

nushea:
### Description
This feature adds basic hiding of rooms and spaces, useful for when taking screenshots of sable if you have private rooms or the like, or if you have unsupportive family members that you do not want to know about your appurtenance to a queer space.

It clears spaces and rooms in the sidebar and the mutual rooms, and is activated per room in each room's appearance settings, or by using Alt-Shift-H for easy access when needed

(before selecting it)
image

(look with it)
image

(it has a setting for toggling it and one for clearing the lists)
image

(before and after in a user card)
image

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

Made with the laziness of me not wanting to clear my screenshots every time

Signed-off-by: Shea <nu@she-a.eu>


src/app/components/GlobalKeyboardShortcuts.tsx
```diff
@@ -75,7 +89,9 @@ export function GlobalKeyboardShortcuts() {
       }
       const roomName = mx.getRoom(roomId)?.name ?? 'Room';
       const roomType = isDirect ? 'Direct Message' : 'Group Room';
-      announce(`${roomName}, ${roomType}. ${remaining} room${remaining === 1 ? '' : 's'} unread.`);
+      announce(
+        `${roomName}, ${roomType}.${remaining && `${remaining} room${remaining === 1 ? '' : 's'} unread.`}`
```
7w1: Won’t this announce “undefined” when remaining is undefined (since it’s optional, and Alt Shift H calls it without remaining)?

src/app/pages/client/sidebar/SpaceTabs.tsx
```diff
@@ -957,7 +960,24 @@ export function SpaceTabs({ scrollRef }: Readonly<SpaceTabsProps>) {
...
+  function handleFilterSpaces(list: SidebarItems, filter: string[] | undefined) {
+    if (filter?.length === 0) return list;
+    let spaces: TSidebarItem[] = [];
+    list.forEach((item) => {
+      if (typeof item === 'object') {
+        const filterSubSpaces = item.content.filter((sId) => !filter?.includes(sId));
+        if (filterSubSpaces.length) {
+          let folder = item;
+          folder.content = filterSubSpaces;
+          spaces.push(folder);
```
7w1, line 973: Is this supposed to mutate or should it be making a copy of `item`? I think mutating might break disabling hiding

src/app/components/GlobalKeyboardShortcuts.tsx
```diff
@@ -32,6 +36,16 @@ export function GlobalKeyboardShortcuts() {
   const roomToUnread = useAtomValue(roomToUnreadAtom);
   const unreadIndexRef = useRef(0);
 
+  const allRooms = useAtomValue(allRoomsAtom);
+  const [isHidingRooms, setIsHidingRooms] = useSetting(settingsAtom, 'isHidingRooms');
+  const [hiddenRooms] = useSetting(settingsAtom, 'hiddenRooms');
+  const [hiddenSpaces] = useSetting(settingsAtom, 'hiddenSpaces');
+  const selectedRoomId = useSelectedRoom();
+
+  const filteredRooms = allRooms.filter(
```
7w1: This should probably be moved inside the hotkey function since it’s only used there, so we don’t recalculate it every time this is called

src/app/pages/client/sidebar/SpaceTabs.tsx
nushea: It should mutate it (to remove the filtered items) but since you should not modify arguments it 'copies' it and pushes into the final version

src/app/components/GlobalKeyboardShortcuts.tsx
nushea: Clarified the logic

src/app/components/GlobalKeyboardShortcuts.tsx
nushea: The hooks should only be recalculated on change as far as i am aware so it shouldnt cause exceeded calculations but it would still make sense to be there, but the issue is that with them being hooks they cannot be put in regular callbacks, or anything that isnt a hook or React function, and reusing a React function would break the style of the other items since it couldnt be called using the useEffect function

7w1: I only meant this line `const filteredRooms = allRooms.filter(...)`. That's not a hook, it should be fine to move that into the callback for the hotkey.

nushea: oh fair that is moved to the proper place

src/app/pages/client/sidebar/SpaceTabs.tsx
7w1: After testing the bug I envisioned doesn't seem to exist so I think its fine.

7w1:
Final thing i noticed, sub spaces don't get hidden

velikopter (changes requested):
I think we should have a list of the hidden rooms which can be unhidden individually instead of needing to unhide all of them.

PR #877, opened 2026-05-19
fix(auth): preserve addAccount param through SSO redirect

Just-Insane:
### Description

When an SSO login is initiated for a second account (with `?addAccount=true`), the parameter was not preserved across the provider redirect and was silently discarded. The returning user was logged in as a new primary session instead of being added to the multi-account switcher. This fix threads the `addAccount` parameter through the SSO callback URL so it survives the redirect.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

The fix reads `addAccount` from the pre-redirect query string and appends it to the `redirectUri` passed to the identity provider, so after the provider redirects back to Sable the login handler adds the session to the multi-account switcher instead of replacing the current session.

### DIFF
--- a/.changeset/auth-sso-addaccount.md
+++ b/.changeset/auth-sso-addaccount.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Preserve addAccount parameter through SSO redirect.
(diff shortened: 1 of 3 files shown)


dozro:
#882 seems to fix this as well 🤷‍♀️

PR #916, opened 2026-06-02
Feat: Add polls

nushea:
### Description

There is a docs written for this feature as https://github.com/SableClient/docs/pull/16

Adds poll events, a poll creation modal, a poll respones modal similar to the reactions one, a /poll command, changes the plus button in the editor to add a menu where the location should be located aswell when implemented, changes the pin behavior to make it so that any click on it takes you there as opposed to needing to hit a small button to align with other platform.

The creation of a new implementation was required as the upstream variant is completely spec-non-compliant, incomplete and broken, so this implementation supersedes it.

(simple snapshot of the change)
image

(examples of the poll creation modal)
image
image

(the new menu in the editor)
image

(the setting to turn it off)
image

(example of look before you cast your vote)
image

(example of undisclosed vote)
image

(example of multiple a choice question)
image

(modal for the voters, chose this size in case that the options become very long)
image
image

 i still so wish more people could test this maybe maybe maybe before merge

I would be very happy to change anything visually about these polls

Fixes #

#### Type of change

- [x] This change requires a documentation update

### Checklist:

### AI disclosure:

Written with the unyielding need of polls as message types

Signed-off-by: Shea <nu@she-a.eu>


src/app/components/message/PollEvent.tsx
```diff
@@ -0,0 +1,336 @@
...
+      roomId,
+      M_POLL_RESPONSE.name as keyof TimelineEvents,
+      newContent as TimelineEvents[keyof TimelineEvents]
+    );
+  }
+  function handleEndVote() {
+    const maxValue = Math.max(...Object.values(votes));
+    let endText = 'The Poll has ended and';
+    const winnerArray = answers.filter((item) => votes[item.id] === maxValue);
+    if (votes.maxValue === 0) endText += ' nobody voted';
```
7w1: `votes.maxValue` should just be maxValue, I think?

7w1 (changes requested):
In the poll itself:
- There should be some more padding between the title, the divider line, and the first item.
- There should be a little more padding between the selector circle/box and the option text.
- The end poll button seems a little large.
Image

In the poll creation screen:
- The options box has scrollbars on chromium based browsers
Image

Below that:
- The options here should maybe have the little gap that we have in settings options, since it looks so similar.
- Also, the draggable selector for the maximum amount of selections seems like its slightly off center?
Image

In general the buttons should probably all use the soft fill that we use for the settings menu for consistency sake.

src/app/components/message/PollEvent.tsx
nushea: corrected because i have only added it later and forgot to apply it there

nushea:
> In the poll itself:
> 
>     * There should be some more padding between the title, the divider line, and the first item.
> 
>     * There should be a little more padding between the selector circle/box and the option text.
> 
>     * The end poll button seems a little large.
> 
> 

Increased the padding but im not sure if it should be any bigger or sth. The button to end poll is as small as the <Button size=""> allows so i could make it smaller but then i would have to force custom sizing
image


> 
> In the poll creation screen:
> 
>     * The options box has scrollbars on chromium based browsers
> 
> 

Fixed this (at least as far as i can tell)

> 
> Below that:
> 
>     * The options here should maybe have the little gap that we have in settings options, since it looks so similar.



>     * Also, the draggable selector for the maximum amount of selections seems like its slightly off center?
> 

it seems to be an issue with how browsers draw ranges as far as i can tell, at least i cant reason why and how its getting not centered

> 
> In general the buttons should probably all use the soft fill that we use for the settings menu for consistency sake.

act changed it now bcs it would probs look better yea

7w1 (approved):
lgtm. styles can be refined further later but i think its good enough to merge

PR #980, opened 2026-06-22
feat: add m.forum room type and dedicated view with threads as topics

smokku:
### Description

This change adds `m.forum` room type support.

Creating room of this type changes the rendering of the main room view, to resemble internet forum post topic, where threaded replies are folded into topic replies.

The room type was chosen (instead room metadata) in order to distinguish the room in the rooms list. Similarly to Voice Rooms.

This feature builds heavily on Lobby and Thread features, combining them to create Forum-like experience. It reuses code where appropriate.

image

image

You can use https://matrix.to/#/!790a18Ov9zgAzyaikc:chrome.pl room to test the feature

#### Type of change

### Checklist:

### AI disclosure:

AI agent was used to update a cherry-picked https://github.com/smokku/cinny/commit/fa94114ddb328223c780de80089f7f293df40752 for Sable codebase and resolve conflicts.


smokku:
@nikiwastaken ☝🏻

src/types/matrix/room.ts
```diff
@@ -20,6 +20,12 @@ export const CustomStateEvent = {
 } as const;
 export type CustomStateEvent = (typeof CustomStateEvent)[keyof typeof CustomStateEvent];
 
+// Custom room types not covered by the Matrix SDK's RoomType enum.
+export const CustomRoomType = {
+  Forum: 'm.forum',
```
dozro: is `m.forum` part of the matrix spec (search for forum on [spec](https://spec.matrix.org/v1.18/client-server-api/) returns nothing at least)?

the `m.` (i.e. `m.xyz`) is reserved for stuff which is part of the spec.

> Vendor prefixes MUST also be used for:
> - New parameters on existing endpoints
> - New properties in existing JSON objects
> - New values for existing parameters or properties

see: https://spec.matrix.org/proposals/#unstable-endpoints-features-and-vendor-prefixes

dozro (changes requested):

src/types/matrix/room.ts
smokku: `m.forum` is not part of the Matrix spec.

dozro: then it should maybe not use the `m.` prefix. in other areas `moe.sable.xyz` is used

see [src/unstable/prefixes/sable/state_events.ts](https://github.com/SableClient/Sable/blob/dev/src/unstable/prefixes/sable/state_events.ts) for example :3

nikiwastaken: smokku/cinny uses m.forum, so maybe we should use `moe.sable.forum` but also support `m.forum`? It just so happened that the only client with forum rooms "support" uses `m.forum`. A forum room is nothing extraordinary anyway, it's just messages and threads.

smokku: If every client supporting forums comes up with own room identifier, it will break other clients, until updated. This will become unwieldy fast.

The protocol recommends using `org.matrix.msc<NUMBER>.<identifier>` so I believe it should be `org.matrix.mscXYZ.forum` with `m.forum` fallback. (We should push for `m.forum` usage by default in the wild, after MSC gets stabilized.)

nikiwastaken: Sure, but for that we need the msc proposed

dozro: > smokku/cinny uses m.forum, so maybe we should use `moe.sable.forum` but also support `m.forum`? It just so happened that the only client with forum rooms "support" uses `m.forum`. A forum room is nothing extraordinary anyway, it's just messages and threads.

you shouldn't use a stable identifier for something which isn't in the spec...

understanding it wouldn't hurt tho, at least if we don't explode if someone else comes around with a completely different `m.forum`

dozro: > If every client supporting forums comes up with own room identifier, it will break other clients, until updated. This will become unwieldy fast.
> 
> 
> 
> The protocol recommends using `org.matrix.msc<NUMBER>.<identifier>` so I believe it should be `org.matrix.mscXYZ.forum` with `m.forum` fallback. (We should push for `m.forum` usage by default in the wild, after MSC gets stabilized.)

that's what MSCs are for, and the proposal process.

the rule with vendor prefixes is to not pollute global namespace with something which might _explode_ other clients

smokku: > Sure, but for that we need the msc proposed

I vaguely remember you offered to help with that. ;-)

smokku: Changed the room type to vendored `pl.chrome.forum`

dozro (approved):
seems fine now (at least the issue i saw before :3)

PR #997, opened 2026-06-26
fix(observability): fix browser profiling + enable CF Worker traces

Just-Insane:
## Summary

- **Browser profiling fix**: `profileLifecycle: 'trace'` was missing from `Sentry.init()`. Without it, the SDK defaults to `'manual'` lifecycle when `profilesSampleRate` is absent (we use `profileSessionSampleRate` for the newer continuous profiling API), meaning the `BrowserProfilingIntegration` never actually starts profiling. Confirmed via `Sentry.getClient().getOptions()` in production — `profileLifecycle` was resolving to `'manual'`.
- **CF Worker observability**: Add `wrangler.jsonc` to enable logs and traces on the `charm` Worker, with `sentry-logs` and `sentry-traces` as export destinations. Head sampling at 100% and persist enabled for both.

## Notes

The `sentry-logs` / `sentry-traces` CF export destinations need to exist as named pipelines in the CF account before forwarding to Sentry works. Traces will still be collected and visible in the CF dashboard regardless.

## Test plan

- [ ] Deploy and verify profiling data appears in Sentry under the `charm` project (Profiling tab)
- [ ] Confirm CF Worker traces appear in the Cloudflare dashboard after deploy
- [ ] Verify `Document-Policy: js-profiling` header still present on production (already confirmed via curl)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Just-Insane:
Opened by mistake, correct PR is CloudHub-Social/Charm#365

PR #1056, opened 2026-07-11
Fix session manager button erroneously appearing when the homeserver does not support it

770grappenmaker:
### Description

Title probably is clear enough. If a homeserver does not report `org.matrix.sessions_list` in `account_management_actions_supported` at the `/_matrix/client/v1/auth_metadata` endpoint, the session manager is not supported, and it shouldn't be assumed to exist just because the hs supports nextgen auth.

#### Type of change

### Checklist:

### AI disclosure:

None of that AI stuff, i am hooman


src/app/features/settings/devices/OtherDevices.tsx
```diff
@@ -116,7 +116,7 @@ export function OtherDevices({ devices, refreshDeviceList, showVerification }: O
       (state: typeof deleteState) => {
         if (state.status === AsyncStatus.Success) {
           setDeleted(new Set());
-          refreshDeviceList();
```
770grappenmaker, line 119: It seems my editor inserted this `void` keyword, which probably does nothing, wat.

bartoostveen: Explicitly ignores the result of `refreshDeviceList` in this case.

src/app/features/settings/devices/OtherDevices.tsx
```diff
@@ -146,11 +146,15 @@ export function OtherDevices({ devices, refreshDeviceList, showVerification }: O
     [authData, deleteDevices, handleCancelAuth]
   );
 
+  // check for MSC4191 support for account deeplink, in particular the sessions list
+  const actionsSupported = authMetadata?.account_management_actions_supported ?? [];
+  const supportsSessionManager = actionsSupported.includes('org.matrix.sessions_list');
```
7w1: `org.matrix.sessions_list` was the unstable prefix, the merged one is `org.matrix.devices_list`, so we should probably check for either of those

770grappenmaker: Aha, I have misunderstood the problem I think, the reason I went about it this way is that Continuwuity does not support `sessions_list`, yet this is explicitly used in the code. I'll fix this, so that if either one is available, that one is picked.

7w1: Tbh I don't actually know exactly what the error you're facing is, but I guess it might have inadvertently been solved by #1060 since i swapped over to the stable keys after seeing your PR, I think all implementations support the merged keys

770grappenmaker:
That should probably do it. Tested, but unsure of the correctness for other HS implementations other than Continuwuity.

770grappenmaker:
Have to admit, not the cleanest thing ever, probably could refactor out generating these actions given metadata.

src/app/features/settings/devices/OtherDevices.tsx
770grappenmaker: In theory the action may not be supported, but indeed your PR solves my issue too

770grappenmaker:
Closing in favor of #1060

PR #1102, opened 2026-07-17
fix: prune left-elsewhere rooms from sliding sync sidebar cache before hydration

eleboucher:
### Description

With sliding sync, the sidebar briefly showed rooms/spaces you'd left on another device, then removed them once reconciliation ran. This fetches the joined-rooms set before hydration and prunes stale rooms from the cache first, so the sidebar renders once at the correct count. Invites are kept; if the request is slow (2s) or offline, it falls back to the old behaviour.

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/.changeset/fix-sidebar-cache-prune-stale-rooms.md
+++ b/.changeset/fix-sidebar-cache-prune-stale-rooms.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix sidebar briefly showing rooms left on another device when loading with sliding sync.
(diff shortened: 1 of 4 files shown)


7w1:
Imo this sort of defeats the purpose of the cache in speeding up load times. For fast homeservers on average account sizes, the difference is negligible and rooms would be cleaned up seconds later regardless. For slow homeservers and large accounts, the full room list would probably take longer than 2 seconds to load anyway, just resulting in the current behavior, meaning all that's changed is an unnecessary 2 second loading screen.

PR #1247, opened 2026-07-23
feat: add multi-channel distribution via GoReleaser

### LINKED ISSUE #379: Tauri - Packaging and Distribution (GoReleaser)
So Tauri out of the box supports a lot of different packaging and even includes some distribution on things like App store, play store and whatnot with help of GitHub actions.

But we still have the native packaging side of the story, especially on Linux. And thing like winget on Windows.

For that I've looked around and the best solution I found that seems to support the most native packaging and distribution is [JReleaser](https://jreleaser.org), originally made for Java and inspired by [GoReleaser](https://goreleaser.com/) but it seems to support other languages like rust now too. This is probably our best bet without having a dedicated system for each package manager. I would've gone with GoReleaser if it didn't paywall some of its features because it seems a lot easier to setup but JReleaser seems to support more things. I've also looked at other options but it would be too many to list, you can ask me if you want to know the other options I've explored.

That being said JReleaser does look very daunting and it seems like we would be the first project to use it for Tauri, so lots of unknowns here.

I will try to take this up myself but I would also really appreciate help here since I don't have any expertise here.

eleboucher:
### Description

GoReleaser now owns Linux packaging.

- `.goreleaser.yaml` builds the CEF binary with the Rust builder and produces deb, rpm and archlinux packages, plus snap and the AUR `sable-bin` package on tagged releases. Nightlies run `release --snapshot`, which packages without publishing, and the workflow uploads those assets. GoReleaser's own `nightly` pipe is Pro-only.
- `nfpm.yaml` is deleted. The `nfpms:` section replaces it, so the per-format dependency list lives in one place.
- `scripts/cef/package.sh` is trimmed to AppImage only. There is no GoReleaser pipe for AppImage, and it is the only Linux format the Tauri updater can update in place. It shares staging with GoReleaser through the new `scripts/cef/stage.sh`.
- `packaging/` holds the AUR `sable-nightly-bin` PKGBUILD and `.install`, plus the deb/rpm postinstall that refreshes the icon and desktop caches.
- `scripts/obtainium-generate.js` emits `obtainium.json` for one-tap Android sideload.
- Linux moves out of the `build` matrix into its own `linux` job, so the CEF binary is built once.

Secrets: `AUR_KEY`, `SNAPCRAFT_STORE_CREDENTIALS`.

#### Type of change

### Checklist:

### AI disclosure:


.github/workflows/tauri-build.yml
```diff
@@ -611,3 +614,183 @@ jobs:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Setup tools (mise)
+        uses: jdx/mise-action@dad1bfd3df957f44999b559dd69dc1671cb4e9ea # v4.2.1
+        with:
+          cache: true
+
+      - name: Install Flatpak build deps (for flatpak builder)
```
hazre: could you get these via mise instead? https://mise.jdx.dev/bootstrap/packages/

.github/workflows/tauri-build.yml
```diff
@@ -611,3 +614,183 @@ jobs:
...
+        env:
+          AUR_KEY: ${{ secrets.AUR_KEY }}
+        run: |
+          mkdir -p ~/.ssh
+          printf '%s\n' "$AUR_KEY" > ~/.ssh/aur_key
+          chmod 600 ~/.ssh/aur_key
+          ssh-keyscan -t ed25519 aur.archlinux.org >> ~/.ssh/known_hosts
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
```
hazre: run mise run actions:pin

.github/workflows/tauri-build.yml
```diff
@@ -611,3 +614,183 @@ jobs:
...
+      VERSION: ${{ needs.setup-release.outputs.version }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Download nightly .deb and compute checksum
+        shell: bash
+        run: |
+          curl -fsSL "https://github.com/${{ github.repository }}/releases/download/${TAG}/Sable-${VERSION}-linux-x86_64.deb" -o sable.deb
```
hazre: can we not use artifacts to pass stuff around instead? https://docs.github.com/en/actions/tutorials/store-and-share-data

packaging/flatpak/moe.sable.Sable.metainfo.xml
hazre, line 1: do we need this? cant we define it within goreleaser config? https://goreleaser.com/customization/package/flatpak/

.goreleaser.yaml
```diff
@@ -0,0 +1,251 @@
...
+  - name: sable
+    publisher: SableClient
+    license: AGPL-3.0-or-later
+    homepage: https://sable.moe
+    short_description: Sable, a Matrix client
+    package_identifier: SableClient.Sable
+    repository:
+      owner: microsoft
+      name: winget-pkgs
+      branch: master
```
hazre: this needs to changed to a fork of winget-pkgs. https://goreleaser.com/customization/publish/winget/#cross-repository-pull-requests

I can make the fork.

.goreleaser.yaml
```diff
@@ -0,0 +1,251 @@
...
+
+changelog:
+  sort: asc
+  use: git
+  filters:
+    exclude:
+      - '^chore:'
+      - '^docs:'
+      - '^style:'
+      - '^test:'
```
hazre: I don't know many in nix community that use NUR. we should publish to nixpkgs instead and maintain our flake. Ideally our flake takes in a nix package. that way we can just PR the package file to nixpkgs.

This is a bit more involved, we can skip it for now if you want.

hazre (changes requested):
also merge conflicts :/

hazre:
Another thing is. I don't like that tauri ci stuff is massive monolith file. can we split it up into multiple workflows/actions? and also its hella messey, we should move any scripting stuff into dedicated js scripts under .github/workflows/scripts to abstract away the complexity.

.goreleaser.yaml
hazre: https://github.com/SableClient/winget-pkgs

hazre:
also ideally sable and nightly setup are identical and we just have a thing that at start of the workflow detects if its a nightly or sable release. to avoid DRY

packaging/flatpak/moe.sable.Sable.metainfo.xml
eleboucher: Dropped the flatpak pipe, so these files are gone. It only builds a `.flatpak` bundle and never submits to Flathub. That needs a manifest in a separate flathub repo.

.goreleaser.yaml
eleboucher: Removed the `nix:` section. Skipping for now as you suggested.

eleboucher:
Line-level comments are addressed. The monolith split and script extraction are out of scope here, worth a follow-up.

.github/workflows/tauri-build.yml
```diff
@@ -637,11 +645,13 @@ jobs:
         uses: jdx/mise-action@dad1bfd3df957f44999b559dd69dc1671cb4e9ea # v4.2.1
         with:
           cache: true
+          env: tauri
 
-      - name: Install Flatpak build deps (for flatpak builder)
+      - name: Install system packages (including Flatpak build deps)
+        run: mise bootstrap packages apply --yes
```
hazre: just use this, already a existing pattern 

```
      - name: Setup app
        uses: ./.github/actions/setup
        with:
          tauri: 'true'
```

hazre (changes requested):

.github/workflows/tauri-build.yml
```diff
@@ -611,3 +622,187 @@ jobs:
...
+          git clone --depth 1 --branch gh-pages \
+            "https://github.com/${{ github.repository }}.git" pages-tmp 2>/dev/null || true
+          if [ -d pages-tmp/fdroid ]; then
+            cp -r pages-tmp/fdroid/. repo/
+          fi
+
+      - name: Download APK from release
+        shell: bash
+        run: |
+          curl -fsSL "https://github.com/${{ github.repository }}/releases/download/${TAG}/Sable-${VERSION}-android-universal.apk" \
```
hazre: Also use artifact pattern here.

.github/workflows/tauri-build.yml
```diff
@@ -611,3 +622,187 @@ jobs:
...
+        shell: bash
+        env:
+          AUR_KEY: ${{ secrets.AUR_KEY }}
+        run: |
+          mkdir -p ~/.ssh
+          printf '%s\n' "$AUR_KEY" > ~/.ssh/aur_key
+          chmod 600 ~/.ssh/aur_key
+          ssh-keyscan -t ed25519 aur.archlinux.org >> ~/.ssh/known_hosts
+
+          git clone "ssh://aur@aur.archlinux.org/sable-nightly-bin.git" aur-repo
```
hazre, line 829: Does this repo exist? How does it work?

hazre (changes requested):

.github/workflows/tauri-build.yml
eleboucher: i guess we will need to create it ourselves

hazre:
conflicts go brrr

hazre (approved):
lgtm, although I doubt it will work first try.

.github/workflows/tauri-build.yml
```diff
@@ -795,3 +795,62 @@ jobs:
...
+        with:
+          persist-credentials: false
+
+      - name: Compute the DMG checksum
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          DMG="Sable-${VERSION}-macos-universal.dmg"
+          gh release download "$TAG" --repo "$GITHUB_REPOSITORY" --pattern "$DMG" --dir dmg
```
hazre: I don't like this. we have access to this information within this workflow.

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs

packaging/flatpak/moe.sable.client.yml
```diff
@@ -0,0 +1,74 @@
...
+      - install -Dm755 sable.sh ${FLATPAK_DEST}/bin/sable
+      - install -Dm644 moe.sable.client.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - install -Dm644 moe.sable.client.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - for s in 32x32 64x64 128x128 256x256; do install -Dm644 "icons/hicolor/$s/apps/sable.png"
+        "${FLATPAK_DEST}/share/icons/hicolor/$s/apps/${FLATPAK_ID}.png"; done
+    sources:
+      - type: archive
+        # The GoReleaser archive has no wrapper directory.
+        strip-components: 0
+        url: https://github.com/SableClient/Sable/releases/download/v1.20.0/Sable-1.20.0-linux-x86_64.tar.gz
```
hazre: this is hardcoded?

hazre (changes requested):

.github/workflows/tauri-build.yml
```diff
@@ -601,3 +559,292 @@ jobs:
...
+        shell: bash
+        run: |
+          if [ "$IS_NIGHTLY" = "true" ]; then
+            echo 'args=release --clean --snapshot --skip=snapcraft,aur' >> "$GITHUB_OUTPUT"
+          else
+            echo 'args=release --clean' >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
```
hazre: I dont think we need this action. we can just use mise to get and install goreleaser instead by adding it to tools in mise.tauri.toml. 

```toml
[tools]
goreleaser = "2.17.0"
```

hazre (changes requested):

packaging/flatpak/moe.sable.client.yml
eleboucher: needed for flathub sorry

hazre (approved):
lgtm

Issue #1279, opened 2026-07-24
🧊 Persona bug/feature request meta-issue


j0lol:
> **Persona work is feature frozen** (and non-urgent bugs are going to be ignored) until 1.21.0. See https://github.com/SableClient/Sable/pull/1110

Personas in Sable are quite experimental and it would be unhelpful to flood the issue tracker with all the things Personas need to get out of experimental. This is an effort to streamline development of Persona/PMP support in Sable and make it highly usable for the usecase of *plural system proxying*.

If you'd like to discuss personas for this use case in depth (or are in the field of plural Matrix tooling,) I have a room at [#plural-wg:f0rest.net](https://matrix.to/#/#plural-wg:f0rest.net). Feel free to join either if you're doing developer work or would like to voice your user perspective and influence design decisions.

## Persona Shorthands 
(Prefixes, Proxy tags, etc)

- [ ] Fix edge cases where shorthands are not used
  - [ ] Cannot send a shorthand with an image
  - [x] Shorthand might not be used if a message contains a newline (Will be fixed by #1352)
- [ ] Make sure shorthand proxy behavior is unambiguous when multiple shorthands could be used
  - Specced in MSCF002
  - Now specced in the PMP storage msc. not implemented (maybe?)
- [x] https://github.com/SableClient/Sable/pull/1307
- [x] **Write tests for proxy tags**
- [x] Support shorthand latching per-room and global (like `pk;autoproxy latch`) https://github.com/SableClient/Sable/pull/1435
  - _Latching_ is a behaviour that adds a room-associated persona after a message with shorthand is sent

## Persona Picker 
(GUI persona switching in room input)

- [ ] Change sort order
- [ ] Create groupings of personas in the picker (e.g. for subsystems, different talksprites)
- [ ] Pin personas to the top of the picker
- [x] Use a popover(?) on mobile, similar to GIF picker (#1368)
- [x] Add global persona tracking (#1264)
- [x] Getting room profile on first render does not load properly. Ameliorated with a hack currently, but may require a fix in `getAllPerMessageProfiles` for better behavior.

## Persona Settings view

- [x] #1369
  -  #1276 
  - Creating a new persona lacks visual feedback (as the operation takes a while to load) 
- [x] #1352 

## PluralKit Compatibility

- [ ] Finish command interface
  - [ ] Add command aliases
  - [ ] Add `pk;switch` for setting global persona
  - [ ] … Many more commands and features…
- [ ] Use a more robust parser
[...]

PR #1352, opened 2026-07-25
Add PMP shorthand editing in Settings

j0lol:
### Description

- Add PMP shorthand editing in Settings
- Migrate from storing shorthands as regex in account data to `{ prefix?, suffix? }`.

image

Part of #1279

#### Type of change

> [!NOTE]  
> This PR is a breaking change on an experimental feature, so I have considered it 'minor' instead of 'major'. Migrations have also been made so users should not expect issues.

### Checklist:

### AI disclosure:

### DIFF
--- a/src/app/features/settings/Persona/PerMessageProfileEditor.tsx
+++ b/src/app/features/settings/Persona/PerMessageProfileEditor.tsx
@@ -1,8 +1,8 @@
 import { SequenceCard, SequenceCardStyle } from '$components/sequence-card';
-import { Box, Button, Text, Avatar, config, IconButton, Input } from 'folds';
-import { menuIcon, X } from '$components/icons/phosphor';
+import { Box, Button, Text, Avatar, config, IconButton, Input, toRem, Spinner, color } from 'folds';
+import { menuIcon, Trash, X } from '$components/icons/phosphor';
 import type { MatrixClient } from '$types/matrix-sdk';
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { nameInitials } from '$utils/common';
 import { mxcUrlToHttp } from '$utils/matrix';
 import { useFilePicker } from '$hooks/useFilePicker';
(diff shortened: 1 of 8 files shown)


j0lol:
really struggling with making this todo list. i'm a fake react dev

src/app/features/settings/Persona/PerMessageProfileEditor.tsx
```diff
@@ -30,9 +39,267 @@ type PerMessageProfileEditorProps = {
...
+  );
+
+  const handleAddShorthand = () => {
+    (shorthands !== undefined) && setShorthands([...shorthands, { id: 'blank' }]);
+  };
+  const constructProxyString = (shorthand: Shorthand) => {
+    return `${shorthand.prefix ?? ''}text${shorthand.suffix ?? ''}`;
+  };
+
+  const [deleteShorthandState, handleDeleteShorthand] = useAsyncCallback(
```
eleboucher, line 187: these close over `shorthands` but the deps are `[mx]`, so the closure keeps the value from the first render, which is `undefined`. the `if (shorthands === undefined) return;` hits every time and nothing can be added, saved or deleted. add `shorthands` to the deps, or just do everything through the `setShorthands(prev => ...)` updater you already use below.

src/app/features/settings/Persona/PerMessageProfileEditor.tsx
```diff
@@ -30,9 +39,267 @@ type PerMessageProfileEditorProps = {
...
+function ShorthandListItem({
+  id,
+  prefix,
+  suffix,
+  onDelete,
+  onSave,
+  deleteState,
+  saveState,
+}: ShorthandListItemProps) {
+  const [currentPrefix] = useState(prefix);
```
eleboucher, line 54: useState(prefix)` which only reads the prop on mount, and the row id changes when you edit it so the row remounts and loses what you typed. give rows a nanoid instead.

j0lol:
Marking this draft quickly while i push through a bugfix for "Shorthand might not be used if a message contains a newline"

j0lol:
fixed (proxy tags now use basic text functions instead of regex which has the potential to break on multiline. also added a test for it)

7w1 (approved):

PR #1367, opened 2026-07-25
Change cross-cutting hooks to shared modules and fix menu, thread and sheet regressions

eleboucher:
### Description

Adds a generalized `useMatrixEvent` hook, `useAsyncAction`/`AsyncButton`, a shared `RoomMenu`/`useRoomMenuActions`, and splits `utils/room.ts` into `unread`, `relations`, `hierarchy` and `display` modules with no barrel. Folds `ThreadDrawer` onto `useTimelineActions` and the shared renderer context, and moves the call feature into `features/call/`.

Also fixes regressions found while reviewing the stack: right-click menus not opening or not anchoring at the cursor, long-press opening the desktop popout instead of the mobile sheet, swipe-to-dismiss shrinking to the drag handle, bottom sheets losing their sheet styling and `role="dialog"`, thread reactions double-reacting instead of redacting, the thread root being unrepliable, and the DM invite prompt never rendering.

#### Type of change

### Checklist:

### AI disclosure:


hazre (approved):
Massive diff but it seems sane to me just looking at the code, I've yet to test it. My only concern is the Timeline and RoomView stuff and making sure there is no regressions from these changes, I'm not familiar with the code there since @7w1 has changed it a lot since last time I looked at it.

hazre:
Long holding a room on mobile in the RoomView just causes it to flash for a second and nothing happens. Is that a regression?

7w1:
> Long holding a room on mobile in the RoomView just causes it to flash for a second and nothing happens. Is that a regression?

Seems like there's something weird going on if you let go too quickly, but I'm only using chromes emulator since holding rooms never worked on iOS anyway. Don't know if its a regression

7w1 (approved):

PR #1373, opened 2026-07-26
Fix the timeline loading a room's entire history on open

### LINKED ISSUE #1371: Opening DMs causes timeline to keep scrolling up
### Describe the bug

See repro

### Reproduction

1. Open any DMs with a decent amount of history
2. Watch it load all of its history, scrolling up in the process

### Expected behavior

It doesn't do that and only loads chat history when I scroll up

### Platform and versions

```shell
1. OS: Android 17 (GrapheneOS v2026071501)
2. Browser: Vanadium WebView v151.0.7922.29.0 (based on Chromium WebView)
3. Sable version: v1.20.1-nightly.2607252113
4. Homeserver: monospacedmagic.dev (running continuwuity:main (commit: 4f1af01bdc) via Podman)
```

### Additional context

_No response_

eleboucher:
### Description

Opening a room paginated until it ran out of history, scrolling the view up as it went. Worst rooms burned 8-10s of CPU per open on a Pixel 3a.

The viewport-fill effect compared `processedEvents.length` against a value captured in the same render, so the guard was always true and the condition reduced to `!hasRealScrollRoom || atTop`. It re-runs on every length change, so staying near the top kept it firing. Pre-existing since March, triggered by #1363 leaving fewer rendered events per page.

Also in here:

- `setAtBottom` was only called from the scroll handler, so a resize (composer, keyboard) left it stale and killed auto-follow. Now recomputed from the ResizeObserver.
- Notification sound gates on window focus instead of tab visibility. A visible but unfocused window is now silent unless background notification sounds is on.
- Desktop reads window focus from `onFocusChanged`, since `focusin`/`focusout` miss app switches. Mobile and web unchanged.

Pixel 3a, release builds, CPU per room open:

| | before | after |
|---|---|---|
| Sable Announcements | 9.72s | 1.06s |
| 1 (DM) | 8.03s | 2.57s |
| 10 DMs | 31.34s | 17.19s |

Fixes #1371

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/.changeset/fix-timeline-late-growth-scroll.md
+++ b/.changeset/fix-timeline-late-growth-scroll.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Keep following the timeline when a message grows after it renders
--- a/.changeset/fix-timeline-runaway-pagination.md
+++ b/.changeset/fix-timeline-runaway-pagination.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Stop the timeline loading a room's whole history on open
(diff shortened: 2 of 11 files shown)


eleboucher:
play wright is flaky because of c10y

7w1 (approved):

PR #1407, opened 2026-07-27
fix(ios): make AltStore/SideStore register nightly updates

eleboucher:
Nightly build version was minute-resolution, so same-minute commits or identical rebuilds produced the same CFBundleVersion. AltStore and SideStore detect updates by exact string-equality on CFBundleVersion, so a collision meant no update was shown. Bump to second resolution.

Add the top-level identifier SideStore requires to link the source to installed apps; AltStore ignores it.

### Description

Fixes #

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/altstore-source.json
+++ b/altstore-source.json
@@ -1,5 +1,6 @@
 {
   "name": "Sable",
+  "identifier": "moe.sable.client.source",
   "subtitle": "Sable for iOS",
   "description": "Sable iOS nightly builds for AltStore and SideStore.",
   "iconURL": "https://raw.githubusercontent.com/SableClient/Sable/dev/src-tauri/icons/icon.png",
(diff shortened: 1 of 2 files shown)


altstore-source.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "Sable",
+  "identifier": "moe.sable.client.source",
```
hazre, line 3: is this correct? and not `moe.sable.client`?

hazre (approved):
Otherwise lgtm.

altstore-source.json
eleboucher: Yes — it's intentional. That top-level `identifier` is SideStore's **source** ID (used to keep sources distinct), not the app's bundle ID. SideStore's type spec describes it as "a bundle identifier SideStore uses to keep your source separate from every other source." AltStore doesn't use it at all.

The app's `bundleIdentifier` stays `moe.sable.client` (unchanged, line 13). Making the source `identifier` the same string as the app's bundle ID would conflate the two namespaces.

The `.source` suffix matches the common convention (e.g. `ai.offgridmobile.source` for the Off Grid source).

eleboucher: it should be diff for altstore

hazre: Okay sounds good

PR #1677, opened 2026-08-07
fix(linux): use a CEF build with H.264 and AAC support

### LINKED ISSUE #1535: MP4 files fail to load
### Describe the bug

I believe this is a codec issue, but on Sable specifically, I cannot see these MP4 files that are being sent

Image

On other clients, such as Element, they load fine

Image

Leaving me to believe this is a client issue.

Both of the MP4 files shown here are AVC codec

### Reproduction

Embed MP4 file (Currently tested as AVC not working)
Try to play
Fails

### Expected behavior

The embedded media should play fine.

### Platform and versions

```shell
Gentoo Linux
Tauri Client v1.20.1-nightly.260730071135-dev (fb87569)
ffmpeg 8.1.2
```

### Additional context

_No response_

eleboucher:
### Description

Fixes #1535

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/scripts/fetch-cef.sh
+++ b/scripts/fetch-cef.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+CEF_VERSION="${CEF_VERSION:-150.0.10+g8042e43+chromium-150.0.7871.101}"
+CEF_RELEASE_URL="${CEF_RELEASE_URL:-https://github.com/tobagin/karere/releases/download/cef-150.0.10-proprietary-codecs}"
+CEF_SHA256_x86_64="3bbe298368c4d87c19ad9b7ed4e8449ea91b32ffa3cefc8672791a1b96c9c3b9"
+CEF_SHA256_aarch64="543bc10ce854fc39b0493ff8b369c47d65fa43f2fb974b734252eea805bffc53"
+
+case "$(uname -m)" in
+  x86_64) PLATFORM=linux64; EXPECTED_SHA="$CEF_SHA256_x86_64" ;;
+  aarch64|arm64) PLATFORM=linuxarm64; EXPECTED_SHA="$CEF_SHA256_aarch64" ;;
+  *) echo "unsupported arch: $(uname -m)" >&2; exit 1 ;;
+esac
+
+CACHE_DIR="${CEF_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/sable-cef}"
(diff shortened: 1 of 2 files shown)


7w1 (changes requested):
I'm not sure we're allowed to do this? 

Proprietary codecs have distribution fees:
https://www.via-la.com/licensing-programs/avc-h-264/#license-fees
https://www.via-la.com/licensing-programs/aac/#license-fees
...which is why most builds don't build with them. If we bundled a cef with those enabled I believe we would be in violation.

On windows/macos it seems like this is worked around by delegating to OS players and whatnot, which seems to be what electron does, wry does similarly since it delegates the entire webview to the system. ~~But for CEF on linux I can't find any actual solution, not sure what electron does there or if it works.~~ Actually electron seems to ship two versions, one with and one without proprietary codecs and just has open issues about licensing? So maybe this is fine? I have no idea

Anyway https://github.com/chromiumembedded/cef/issues/3559 seems to imply we would be subject to licensing fees

eleboucher:
It's only a cef issue not a wry because WebKit have the proprietary stuff (same as electron so I guess it's fine)

Bu156:
I'm not that familiar with CEF, but could it not use the system's ffmpeg installation for codecs? Then you wouldn't need to distribute them at all, is my thinking.

eleboucher:
> I'm not that familiar with CEF, but could it not use the system's ffmpeg installation for codecs? Then you wouldn't need to distribute them at all, is my thinking.

it's possible but then we would need to build cef ourselves, i will see if i can find it online as a deps

eleboucher:
> I'm not sure we're allowed to do this?
> 
> Proprietary codecs have distribution fees: https://www.via-la.com/licensing-programs/avc-h-264/#license-fees https://www.via-la.com/licensing-programs/aac/#license-fees ...which is why most builds don't build with them. If we bundled a cef with those enabled I believe we would be in violation.
> 
> On windows/macos it seems like this is worked around by delegating to OS players and whatnot, which seems to be what electron does, wry does similarly since it delegates the entire webview to the system. ~But for CEF on linux I can't find any actual solution, not sure what electron does there or if it works.~ Actually electron seems to ship two versions, one with and one without proprietary codecs and just has open issues about licensing? So maybe this is fine? I have no idea
> 
> Anyway [chromiumembedded/cef#3559](https://github.com/chromiumembedded/cef/issues/3559) seems to imply we would be subject to licensing fees

> The OS manufacturers (Microsoft and Apple respectively) allow applications to use these OS platform APIs with no added licensing requirements or costs.

so we are actually good

PR #1779, opened 2026-08-13
fix(timeline): restore hidden event ordering

7w1:
### Description

Fixes a regression caused sometime in the recent timeline fixes that caused hidden events like reactions to be placed inaccurately in the timeline, usually bundled right behind their parent instead of at the actual time

#### Type of change

### Checklist:

### AI disclosure:

### DIFF
--- a/src/app/hooks/timeline/useProcessedTimeline.test.tsx
+++ b/src/app/hooks/timeline/useProcessedTimeline.test.tsx
@@ -35,6 +35,7 @@ type FakeEventOptions = {
   relation?: { rel_type: string; event_id: string; key?: string };
   threadRootId?: string;
   ts?: number;
+  localTimestamp?: number;
   isRedaction?: boolean;
 };
 
@@ -47,6 +48,7 @@ function createEvent({
   relation,
   threadRootId,
   ts = 1_000_000,
+  localTimestamp = ts,
   isRedaction = false,
 }: FakeEventOptions): MatrixEvent {
   return {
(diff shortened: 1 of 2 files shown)


eleboucher (changes requested):
No it's agaisn't the spec

src/app/hooks/timeline/useProcessedTimeline.test.tsx
```diff
@@ -35,6 +35,7 @@ type FakeEventOptions = {
   relation?: { rel_type: string; event_id: string; key?: string };
   threadRootId?: string;
   ts?: number;
+  localTimestamp?: number;
```
eleboucher, line 38: hsTimestamp is better

src/app/hooks/timeline/useProcessedTimeline.ts
```diff
@@ -258,7 +261,14 @@ const mergeDraftsAndExtras = (
       eventSender: mEvent.getSender() ?? null,
     },
     parentId,
+    timelineTs: getTimelineTimestamp(mEvent),
   }));
+  const orderedExtraDrafts = [
```
eleboucher: Sorting inline reactions before fetched ones can still render an older fetched reaction after a newer inline reaction; sort both together.

src/app/hooks/timeline/useProcessedTimeline.test.tsx
7w1: hsTimestamp isn't provided by the sdk which is why this fallback is necessary

eleboucher (approved):

PR #1972, opened 2026-08-27
fix(shortcuts): make the global show/hide window shortcut opt-in and configurable

### LINKED ISSUE #1971: Sable takes over Ctrl+Shift+S hotkey globally
### Describe the bug

Pressing `Ctrl`+`Shift`+`S` in any application (not Sable) focuses Sable window instead of passing the hotkey to the currently focused application.
I couldn't find any related settings to disable this.
A lot of applications use this hotkey for "Save as" or "Save all" action and Sable makes it impossible to use.

### Reproduction

1. Open Sable
2. Open any other application and make sure it's focused (for example, Notepad)
3. Press `Ctrl`+`Shift`+`S`

### Expected behavior

The shortcut gets passed to the currently focused window

### Platform and versions

```shell
OS: Windows 11
Sable Version: v1.21.0-dev (7675aa1) (installed via the MSI package).
```

### Additional context

_No response_

Th-Underscore:
Closes #1971.

### Description

Sable currently registers `Ctrl+Shift+S` as a system-wide show/hide hotkey on startup, unconditionally and non-configurably - hijacking the key from every other application (it collides with "Save As" in most of them). This makes the shortcut:

- **Opt-in**: nothing is registered until the user assigns a binding. Absent/null/empty settings all mean "unregistered", and Sable no longer claims any global key on its own.
- **Configurable**: a new `Show/Hide Sable` entry in Settings -> Keyboard Shortcuts. Assigning a binding registers it with the OS _first_ - a binding the OS rejects is never persisted - and clearing it (Backspace/Delete) unregisters immediately. Changes apply live; startup re-registers whatever was saved.
- Plumbing: a new `set_global_window_shortcut` Tauri command persists to the desktop settings store and re-registers, `menu.rs` reads the binding from the store on boot, and a `toTauriAccelerator` converter maps captured web bindings to Tauri accelerator strings.

Tested on the desktop app, Ubuntu 24.04, as well as Windows 11 when I first built this a couple weeks ago.

One thought that's deliberately out of scope for this PR: since the shortcut is now off by default, new users have no natural way to discover it. It may be worth a one-time notification or a hint somewhere pointing at Settings -> Keyboard Shortcuts, so the feature doesn't become invisible to everyone who'd actually want it.

#### Type of change

- [X] New feature (non-breaking change which adds functionality)

### Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation - `.changeset`

### AI disclosure

- [ ] Partially AI assisted
- [X] Fully AI generated (explain what all the generated code does in moderate detail)

I hit this bug myself and tested the builds; the code was AI-generated. The Rust side stores the binding under the existing desktop-settings store (`globalWindowShortcut` key), `register_global_shortcuts` treats missing/null/empty as unregistered, and the command unregisters-all then re-registers the new binding. The frontend adds the shortcut definition with a `null` default, the capture flow routes through the command before persisting, and an effect keeps the OS registration in sync with the persisted value.

### DIFF
--- a/src/app/keyboard/shortcuts.ts
+++ b/src/app/keyboard/shortcuts.ts
@@ -9,6 +9,7 @@ export type ShortcutId =
   | 'app.searchMessages'
   | 'app.openBookmarks'
   | 'app.createRoom'
+  | 'app.toggleWindowVisibility'
   | 'navigation.nextUnread'
   | 'navigation.cycleNextUnread'
   | 'navigation.cyclePreviousUnread'
@@ -38,7 +39,7 @@ export type ShortcutDefinition = {
   id: ShortcutId;
   label: string;
   category: 'General' | 'Navigation' | 'Messages';
-  defaultBinding: string;
+  defaultBinding: string | null;
   scope: ShortcutScope;
   allowInEditable?: boolean;
 };
@@ -64,6 +65,14 @@ export const SHORTCUTS: readonly ShortcutDefinition[] = [
     label: 'Create a room',
     category: 'General',
     defaultBinding: 'mod+shift+n',
+    scope: 'global',
+  },
+  {
+    id: 'app.toggleWindowVisibility',
+    label: 'Show/Hide Sable',
+    category: 'General',
+    // Opt-in: disabled by default so Sable never claims a system-wide key on its own.
+    defaultBinding: null,
     scope: 'global',
   },
   {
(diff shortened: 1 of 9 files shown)


src/app/keyboard/shortcuts.ts
```diff
@@ -294,6 +303,29 @@ export const captureShortcut = (event: ShortcutEvent): string | undefined => {
   return [...modifiers, key].join('+');
 };
 
+export const toTauriAccelerator = (binding: string): string => {
```
lunar-seal: A comment explaining this would be good. Why is this needed? What kind of keycodes are we translating from, what are we translating to?

src/app/keyboard/shortcuts.ts
```diff
@@ -66,6 +67,14 @@ export const SHORTCUTS: readonly ShortcutDefinition[] = [
     defaultBinding: 'mod+shift+n',
     scope: 'global',
   },
+  {
```
lunar-seal: This setting is currently being synced across all devices, and shown on every platform.

eleboucher (approved):

